PR #89, opened 2020-06-01
Fix `self-update` command. Refactor container and telemetry implementation.

grasmash:


bin/acli
```diff
@@ -53,29 +57,13 @@ $output = new ConsoleOutput();
 $logger = new ConsoleLogger($output);
 $container = new Symfony\Component\DependencyInjection\ContainerBuilder();
 $container->setParameter('repo_root', find_repo_root());
-$container->set('amplitude', Amplitude::getInstance());
```
grasmash, line 56: @danepowell There was too much duplication here and in the TestBase class, so I've moved much of this to a static method that can be shared.

src/AcquiaCliApplication.php
```diff
@@ -78,10 +83,7 @@ public function __construct(
     $this->warnIfXdebugLoaded();
     $this->setSshHelper(new SshHelper($this, $output));
     parent::__construct('acli', $version);
-    $definition = $container->register('cloud_api', ClientService::class);
-    $definition->setArgument(0, $this->getContainer()->get('cloud_datastore'));
```
grasmash, line 82: @danepowell I've replaced this with a reference to the service elsewhere.

src/AcquiaCliApplication.php
```diff
@@ -107,131 +109,57 @@ public function __construct(
...
+
+    $container->register('acli_datastore', JsonFileStore::class)
+      ->addArgument($container->getParameter('acli_config.filepath'));
+
+    $container->register('cloud_datastore', JsonFileStore::class)
+      ->addArgument($container->getParameter('cloud_config.filepath'))
+      ->addArgument(JsonFileStore::NO_SERIALIZE_STRINGS);
+
+    $container->register('cloud_api', ClientService::class)
+      ->addArgument(new Reference('cloud_datastore'));
```
grasmash, line 139: @danepowell Example of reference to another service.

src/Command/CommandBase.php
```diff
@@ -71,38 +74,46 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
...
 
+
   /**
    * Check if telemetry preference is set, prompt if not.
    */
-  protected function checkTelemetryPreference() {
-    $datastore = $this->getApplication()->getContainer()->get('acli_datastore');
-    $telemetry = $datastore->get(DataStoreContract::SEND_TELEMETRY);
-    if (!isset($telemetry) && $this->input->isInteractive()) {
+  public function checkTelemetryPreference(): void {
```
grasmash, line 109: I'd prefer for this to be in the TelemetryHelper class, but for some reason doing that causes the $inputs for the CommandTester to not register. Something weird about dependency injection of inputs?

src/Command/CommandBase.php
```diff
@@ -71,38 +74,45 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
...
    * @throws \Acquia\Cli\Exception\AcquiaCliException
+   * @throws \Exception
    */
   protected function initialize(InputInterface $input, OutputInterface $output) {
     $this->input = $input;
     $this->output = $output;
     $this->formatter = $this->getHelper('formatter');
     $this->setLogger(new ConsoleLogger($output));
+
+    /** @var TelemetryHelper $telemetry_helper */
```
grasmash, line 85: @danepowell I'd love to find a way to avoid needing to do this.

tests/phpunit/src/TestBase.php
```diff
@@ -94,38 +106,39 @@ protected function setUp($output = NULL): void {
...
     $this->removeMockConfigFiles();
     $this->createMockConfigFile();
 
     parent::setUp();
   }
 
+  /**
+   * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
+   */
+  protected function configureContainer(ContainerBuilder $container): void {
```
grasmash, line 130: @danepowell The refactor let me get away with overriding far fewer things here.

tests/phpunit/src/TestBase.php
```diff
@@ -268,10 +281,16 @@ protected function createLocalSshKey($contents): string {
   }
 
   protected function createMockConfigFile(): void {
-    $contents = json_encode(['key' => 'testkey', 'secret' => 'test']);
+    // @todo Read from config object.
+    $default_values = ['key' => 'testkey', 'secret' => 'test'];
+    $cloud_config = array_merge($default_values, $this->cloudConfig);
```
grasmash, line 286: @danepowell Note that this lets us extend the default config easily in a test.

src/Command/CommandBase.php
danepowell: Yeah I think this will all become much cleaner if/when we can inject dependencies via Symfony service config, instead of constantly polling the container.

src/Command/CommandBase.php
grasmash: @danepowell I'd like to have a follow up PR to move this to TelemetryHelper. We can debug there.

src/Helpers/TelemetryHelper.php
```diff
@@ -0,0 +1,157 @@
...
+      return;
+    }
+    try {
+      $amplitude->init('956516c74386447a3148c2cc36013ac3');
+      // Method chaining breaks Prophecy?
+      // @see https://github.com/phpspec/prophecy/issues/25
+      $amplitude->setDeviceId(OsInfo::uuid());
+      $amplitude->setUserProperties($this->getTelemetryUserData($app_version));
+      $amplitude->setUserId($this->getUserId());
+      $amplitude->logQueuedEvents();
```
grasmash, line 75: @danepowell Note that I'm only logging events if the user has opted into sending telemetry.

src/Helpers/TelemetryHelper.php
```diff
@@ -0,0 +1,157 @@
...
+    } catch (IdentityProviderException $e) {
+      // If something is wrong with the Cloud API client, don't bother users.
```
grasmash, line 77: I'm on the fence about whether to log an error here.

danepowell: I chose not to log an error because any negative experience with telemetry (even a warning message in logs) could invite people to just stop using it. And if there's really a problem with the Cloud API client, it will probably be revealed in mainline commands that will throw an error anyway.

PR #93, opened 2020-06-02
Move checkTelemetryPreference() to TelemetryHelper.

grasmash:


grasmash:
@danepowell Any thoughts on why this would fail? It seems that the tests $inputs no longer are passed to the $input when this change is made.

grasmash:
I've got it diagnosed. It's an issue that will only occur during a test. Since we first create a placeholder $input and then configure the container with that and we LATER pass arguments to execute the command, the arguments never actually propagate to the $input that is used by the container. There are effectively 2 different $inputs.

That doesn't happen with actual execution because we already know the arguments when the command is executed, and we use only a single $input object to configure the container and subsequently run the command.

tests/phpunit/src/CommandTestBase.php
```diff
@@ -74,14 +69,16 @@ protected function setCommand(Command $command): void {
...
    */
   protected function executeCommand(array $args = [], array $inputs = []): void {
     $cwd = $this->projectFixtureDir;
     chdir($cwd);
-    $tester = $this->getCommandTester();
-    $tester->setInputs($inputs);
-    $command_name = $this->command->getName();
-    $args = array_merge(['command' => $command_name], $args);
+
+    $input = new ArrayInput($args);
```
grasmash, line 78: The idea of this refactor is to delay creation of $input until just before the command is executed so that our container and our executed command both use exactly the same input object.

tests/phpunit/src/TestBase.php
```diff
@@ -85,37 +84,50 @@ abstract class TestBase extends TestCase {
...
     $this->amplitudeProphecy = $this->prophet->prophesize(Amplitude::class);
     $this->clientProphecy = $this->prophet->prophesize(Client::class);
 
-    $container = new ContainerBuilder();
-    $container->setParameter('repo_root', $this->projectFixtureDir);
-    $container->setParameter('data_dir', $this->fixtureDir . '/.acquia');
-    AcquiaCliApplication::configureContainer($container, $this->input, $output, $logger);
-    $this->configureContainer($container);
-
-    $this->application = new AcquiaCliApplication($container, $logger, $this->input, $output, 'UNKNOWN');
```
grasmash, line 117: We delay building the container and creating the application until just before command execution.

danepowell:
I just discovered ORCA also uses Zumba/Amplitude, I wonder if we can steal--er, borrow, or share--any components from ORCA to get this working and streamlined.

I'm also trying to rewire the whole app using symfony autowiring using ORCA as a template.

grasmash:
Assuming your work to refactor and autowire enables us to avoid instantiating $input ourselves during test set up, I'm going to pause work on this PR.

tests/phpunit/src/TestBase.php
```diff
@@ -85,37 +84,50 @@ abstract class TestBase extends TestCase {
...
   protected function setUp($output = NULL): void {
     if (!$output) {
       $output = new BufferedOutput();
     }
+    $this->applicationOutput = $output;
 
     $this->fs = new Filesystem();
     $this->prophet = new Prophet();
     $this->consoleOutput = new ConsoleOutput();
-    $this->input = new ArrayInput([]);
```
grasmash, line 104: @danepowell This is what we need to stop doing.

PR #94, opened 2020-06-03
Use dependency injection

danepowell:
- [x] Figure out how to warm the Symfony cache when building phar files. [Ordinarily we’d call](https://github.com/humbug/box/blob/master/doc/symfony.md) `cache:clear` as part of composer autoload dump and this would warm the cache, but that doesn’t work since we aren’t using framework bundle. That's why ORCA reimplemented `cache:clear` as a pseudo-command but not with support for warming. Running `./bin/acli` might be sufficient.
- [x] Clean up remaining commands to stop accessing container (sed replace)
- [x] For each command, add static name and restore disabled tests (one at a time but use sed to clean up others simultaneously)
- [x] Clean up services.yml
- [x] Kill AcquiaCliApplication


bin/acli.php
```diff
@@ -0,0 +1,61 @@
...
+use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Debug\Debug;
+use Symfony\Component\Filesystem\Filesystem;
+
+if (!file_exists(__DIR__ . '/../vendor/autoload.php')) {
+  die("Could not find autoloader. Run 'composer install' first.\n");
+}
+require __DIR__ . '/../vendor/autoload.php';
+
+ini_set('memory_limit', -1);
```
grasmash, line 21: Eek!

bin/acli.php
```diff
@@ -0,0 +1,61 @@
...
+    // phpcs:enable
+  }
+}
+
+$kernel = new Kernel($env, $debug);
+
+// Handle a cache:clear pseudo command. This isn't implemented as a true console
+// command because a stale or corrupted cache would render it unusable--
+// precisely when it is needed.
+if (in_array($input->getFirstArgument(), ['cache:clear', 'cc'])) {
```
grasmash, line 43: Interesting. I think we may be actually caching in OpsCache. Will this clear that? 
https://github.com/acquia/cli/blob/master/src/Command/Api/ApiCommandHelper.php#L27

config/services.yml
```diff
@@ -0,0 +1,49 @@
...
+  Symfony\Component\Console\Application: ~
+
+  Symfony\Component\Console\Input\ArgvInput: ~
+
+  Symfony\Component\Console\Input\InputInterface:
+    alias: Symfony\Component\Console\Input\ArgvInput
+
+  Symfony\Component\Console\Logger\ConsoleLogger: ~
+
+  Symfony\Component\Console\Output\ConsoleOutput: ~
```
grasmash, line 35: Can we add some comments to explain what's happening on some of these lines?

src/Command/AuthCommand.php
```diff
@@ -14,20 +16,54 @@
...
  */
-class AuthCommand extends CommandBase {
+class AuthCommand extends Command {
+
+  /**
+   * The default command name.
+   *
+   * @var string
+   */
+  protected static $defaultName = 'auth:login';
```
grasmash, line 31: Why do this statically rather than in the configure() method?

src/Helpers/LocalMachineHelper.php
```diff
@@ -239,6 +240,10 @@ public static function getHomeDir(): string {
     return $home;
   }
 
+  public static function getCloudConfigFilepath(): string {
+    return self::getHomeDir() . '/.acquia/cloud_api.conf';
```
grasmash, line 244: We may need to define a constant for `'/.acquia/cloud_api.conf'` so that we can easily set a different file path during testing. Maybe you've already solved that, though.

bin/acli.php
danepowell: ORCA does this, I assumed there's a good reason, but we don't have too.

bin/acli.php
danepowell: I don't think this would clear the api cache. Symfony caches service config in `{repo.root}/var/cache`, only this directory gets cleared when you run `cache:clear`. And it's only useful during development when you are changing services, I don't think an end user would ever need it.

src/Command/AuthCommand.php
danepowell: Doing this statically is required to [support lazy-loading](https://symfony.com/doc/current/console/commands_as_services.html#console-command-service-lazy-loading). I'm not sure why you'd want to use the `configure()` method, maybe that was a best practice in earlier Symfony versions.

config/services.yml
danepowell: These are all just 'services' (specific classes) that are part of some constructor in our application. Symfony probably _could_ autodiscover all of these, but it's way more performant (at least for vendor-provided services) to enumerating, that's what we're doing here.

I think we should only comment on ones that might be non-obvious, anything in particular bother you?

src/Helpers/LocalMachineHelper.php
danepowell: I added some parameters to services.yml, including `app.data_dir`, which is I think what you want. We already inject a custom directory for testing.

config/services.yml
danepowell: I added some comments.

danepowell:
#99

PR #99, opened 2020-06-04
Use dependency injection

danepowell:
There's definitely room for optimization here, but it works and I think it would be good to get this merged to avoid conflicts, since it touches pretty much every file.

The diff is alarmingly large, but if you subtract lockfile changes the LOC count actually goes down, so this does somewhat simplify the codebase. And more importantly, it brings us up from the Symfony 2-3ish design era to Symfony 5+ standards.

If you're not familiar with Symfony service config: the first time you run `acli` it builds a command cache in the `var` directory, which takes a few hundred ms, but on subsequent runs it's blazing fast. When we package the phar we will include that cache so end users have a great experience.

The main opportunity for improvement here is around the test cases, simplifying how the dependencies get injected. And related to that, maybe don't inject every possible dependency for the base command. Also it would be good to attach the event dispatcher inside of the application instead of in the binary.


.env
```diff
@@ -0,0 +1,16 @@
...
+#  * .env.$APP_ENV       committed environment-specific defaults
+#  * .env.$APP_ENV.local uncommitted environment-specific overrides
+#
+# Real environment variables win over .env files.
+#
+# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
+#
+# Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
+# https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration
+APP_ENV=prod
```
grasmash, line 15: Do we really want prod to be the default? How is this used?

config/services.yml
```diff
@@ -0,0 +1,70 @@
+---
+parameters:
+  env(ACLI_DATA_DIR): "%env(HOME)%"
+  app.data_dir: "%env(HOME)%/.acquia"
```
grasmash, line 4: Nice!

config/services.yml
```diff
@@ -0,0 +1,70 @@
...
+    bind:
+      $cloudConfigFilepath: '%app.cloud_config_filepath%'
+      $acliConfigFilename: '%app.acli_config_filename%'
+      $repoRoot: "%kernel.project_dir%"
+      $sshDir: "%app.ssh_dir%"
+      Webmozart\KeyValueStore\JsonFileStore $datastoreCloud: '@datastore.cloud'
+      Webmozart\KeyValueStore\JsonFileStore $datastoreAcli: '@datastore.acli'
+    public: true
+
+  # Register nearly all Acquia CLI classes as services.
```
grasmash, line 25: Why would a command be a service? Is that required for injecting a dependency into it?

src/Command/Api/ApiCommandBase.php
```diff
@@ -41,7 +41,7 @@ class ApiCommandBase extends CommandBase {
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
     // Build query from non-null options.
-    $acquia_cloud_client = $this->getApplication()->getContainer()->get('cloud_api')->getClient();
+    $acquia_cloud_client = $this->clientService->getClient();
```
grasmash, line 44: I'd like to name this something like cloudApiService, since you could easily have clients for other APIs like Pipelines API.

.env
danepowell: It's a little confusing, I know. Basically it doesn't matter what we call the environment, but we have to call it _something_, and Symfony defaults to "prod". Symfony does nothing with it internally. Unless we go out of our way to e.g. change configuration based on environment, it simply doesn't matter.

The second parameter, debug, _does_ matter. That should be `false` when we ship, but `true` when debugging locally.

https://symfony.com/doc/4.1/configuration/environments.html

config/services.yml
danepowell: In Symfony 5, every command is a service. It's required for dependency injection.

> If you're using the default services.yaml configuration, your command classes are already registered as services. Great! This is the recommended setup.


https://symfony.com/doc/current/console/commands_as_services.html

symfony.lock
```diff
@@ -0,0 +1,380 @@
+{
```
grasmash, line 1: What is a Symphony.lock for?

danepowell: It's part of Flex, which provides Composer utilities to go along with Symfony. It's kind of cool actually, packages (like PHPUnit) can provide Flex recipes that update files on disk (such as tests/bootstrap.php).

symfony.lock is what Flex uses to keep track of what recipe versions have been applied.

PR #108, opened 2020-06-05
Move checkTelemetryPreference() to TelemetryHelper.

grasmash:


tests/phpunit/src/TestBase.php
```diff
@@ -162,8 +157,7 @@ protected function setUp($output = NULL): void {
     $this->fs = new Filesystem();
     $this->prophet = new Prophet();
     $this->consoleOutput = new ConsoleOutput();
-    $this->input = new ArrayInput([]);
```
grasmash, line 165: @danepowell Do you have a suggestion about what to do here? We need to avoid defining both $input and $output until we're ready to execute the command. But the LocalMachineHelper, TelemetryHelper, and SshHelper services need these in order to be instantiated.

We could just mock both services _all the time_ like you do for the clientServiceProphecy, but I'm not excited about all the mock calls we need to support in that case. 

Can't we use the container to defer this somehow?

Until we do, our helper will be using different $input and $output objects than our commands during testing.

danepowell: Our unit tests don't boot the kernel or build the container, so every service is "mocked" in some sense, whether it uses the real class (as with SshHelper) or a double (prophecies).

We could _probably_ boot the kernel for unit tests, but it might be another decently large refactor. It would simplify some things because we wouldn't have to mock as much, but I don't have a sense for the LOE or whether it's the "right" thing to do.

Note that Symfony seems to recommend running unit tests with CommandTester (as we do) using KernelTestCase, which would boot the kernel and compile the container. Unfortunately we can't use KernelTestCase because it depends on the Framework Bundle, which is a huge web-oriented dependency that we probably don't want. So we'd have to reinvent the wheel, same as we did for the cache:clear command.

PR #113, opened 2020-06-05
Remove all cache dirs with acli cc.

grasmash:


danepowell:
@TravisCarden you may be interested in porting this to ORCA, since we ~stole~ borrowed our binary from ORCA and maybe you've run into issues related to stale caches when running ORCA.

danepowell:
Oh nevermind, now I see a little more of what's going on here. This only happens because we use the `PhpArrayAdapter`. ORCA doesn't use that, so it's not affected.

PR #125, opened 2020-06-09
Removing command cache.

grasmash:


danepowell:
This adds 25-50ms to all commands. It's not a huge problem, but it's not ideal.

It also still uses the system pool cache (/tmp) for the api spec. Granted, that changes less frequently, but it still could cause problems.

The ideal solution that would eliminate the system cache entirely and preserve speed would be to fix our usage of PhpArrayAdapter to use `warmUp()` rather than `set()`.

danepowell:
Also I think this is needlessly rewriting the cache on every call, regardless of whether the cache is stale or not. That could account for the speed difference.

danepowell:
@grasmash I pushed a change to this PR to use the `var` cache, so that it can be shipped with the phar and there's no risk of having a stale user cache. There's still a very minor performance penalty by removing the command cache, but this seems like a good compromise.

PR #130, opened 2020-06-09
Checking blt.yml for Cloud app uuid.

grasmash:


src/Command/CommandBase.php
```diff
@@ -596,6 +598,15 @@ protected function doDetermineCloudApplication() {
       return $application_uuid;
     }
 
+    // Try blt.yml.
+    $blt_yaml_file_path = Path::join($this->repoRoot, 'blt', 'blt.yml');
+    if (file_exists($blt_yaml_file_path)) {
+      $contents = Yaml::parseFile($blt_yaml_file_path);
+      if (array_key_exists('cloud', $contents) && array_key_exists('appId', $contents['cloud'])) {
+        return $contents['cloud']['appId'];
```
grasmash, line 606: @danepowell Can you verify this is the correct key?

danepowell: Yep that's correct. Although just to be clear, BLT never actually uses this key, it's only ever set/read by ADS CLI right now. Still useful for folks migrating from ADS CLI though...

grasmash: Yeah this is just for backwards support.

PR #143, opened 2020-06-10
Fixes #142: ide:wizard:ssh-key:create-upload broken due to missing sh…

### LINKED ISSUE #142: ide:wizard:ssh-key:create-upload broken due to missing shell script
Attempting to run `ide:wizard:ssh-key:create-upload` will throw an error:
```
In IdeWizardCreateSshKeyCommand.php line 126:
  Unable to add SSH key to local SSH agent:ssh_askpass: exec(phar:///usr/local/bin/acli/src/Command/Ide/Wizard/passphrase_prompt.sh)  
  : No such file or directory
```

We thought we fixed this by including the shell script in the phar, but I confirmed it's there and we still get the error.

I think the problem is that there's simply no way to run a shell script packaged inside a phar, without doing some gymnastics: https://stackoverflow.com/questions/41710814/execute-helper-script-inside-of-phar

grasmash:
…ell script.

### DIFF
--- a/src/Command/Ide/Wizard/IdeWizardCreateSshKeyCommand.php
+++ b/src/Command/Ide/Wizard/IdeWizardCreateSshKeyCommand.php
@@ -126,6 +126,7 @@ class IdeWizardCreateSshKeyCommand extends IdeWizardCommandBase {
 echo $SSH_PASS
 EOT
     );
+    $this->localMachineHelper->getFilesystem()->chmod($temp_filepath, 0755);
 
     $private_key_filepath = str_replace('.pub', '', $filepath);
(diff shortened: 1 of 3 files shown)


src/Command/Ide/Wizard/IdeWizardCreateSshKeyCommand.php
```diff
@@ -119,9 +119,18 @@ protected function localIdeSshKeyExists(): bool {
    */
   protected function addSshKeyToAgent($filepath, $password): void {
     // We must use a separate script to mimic user input due to the limitations of the `ssh-add` command.
-    $passphrase_prompt_script = __DIR__ . '/passphrase_prompt.sh';
+    // @see https://www.linux.com/topic/networking/manage-ssh-key-file-passphrase/
+    $temp_filepath = $this->localMachineHelper->getFilesystem()->tempnam(sys_get_temp_dir(), 'acli');
+    $this->localMachineHelper->writeFile($temp_filepath, <<<'EOT'
```
grasmash, line 124: I'm not too worried about this given that this command is intended only to be run inside of an IDE.

PR #144, opened 2020-06-10
Always use highest versions of dependencies.

grasmash:


.travis.yml
```diff
@@ -62,7 +58,7 @@ before_install:
 install:
   # Load composer dependencies.
   - composer validate --no-check-all --ansi
-  - 'composer -n ${HIGHEST_LOWEST-install} --prefer-dist -o'
+  - 'composer -n update --prefer-dist -o'
```
danepowell, line 61: We should just `composer install` here, not `composer update`.

grasmash: Why? Shouldn't we be apprised of upstream issues that might break the project?

danepowell: It's one thing to test with updated versions, but this is actually building the project with brand new and unknown versions. That seems risky.

I think it's better to have fixed and predictable dependency versions (via Composer.lock), plus that makes for much faster builds. We should update those dependencies on some regular basis, but not automatically on every build.

PR #151, opened 2020-06-11
Fixes #150: acli refresh dumps the MySQL file in the project root and…

### LINKED ISSUE #150: acli refresh dumps the MySQL file in the project root and doesn't clean it up
After a successful `acli refresh`, the SQL dump remains in the project root. We should ensure it's being cleaned up.

grasmash:
… doesn't clean it up.


anavarre:
FYI this fix didn't work. I'm still seeing the DB dump in the project root with beta7.

PR #157, opened 2020-06-12
Cannot run on Windows

danepowell:
This is just a start. It converts the sshDir, but we need to convert nearly _all_ of the parameters (anything that depends on `$_ENV['HOME']`). I'm not sure how to feel about this approach. Reducing the number of service parameters / dependencies injected is good I think. Feels weird to kill all parameter though.

The reason this is so complicated: we use the home directory in our parameters to set other things like the data and ssh dirs, but HOME isn't set on Windows, it's some other variable. You can't use expressions in parameters, they have to be static, so we can't dynamically set the home directory.

I think we'll need to strip out most of the parameters and convert them into methods on the LocalMachineHelper, and then pass the LocalMachineHelper as a service and call its methods to get the values instead of getting them as parameters.

Alternatively, we might be able to bind these paths instead of setting them as parameters. That would involve a lot less refactoring. But the static methods can't be tested as easily, we'd need new tests.

### DIFF
--- a/config/services.yml
+++ b/config/services.yml
@@ -3,7 +3,6 @@ parameters:
   env(ACLI_REPO_ROOT): "%kernel.project_dir%"
   app.data_dir: "%env(HOME)%/.acquia"
   app.repo_root: "%env(ACLI_REPO_ROOT)%"
-  app.ssh_dir: "%env(HOME)%/.ssh"
   app.acli_config_filename: 'acquia-cli.json'
   app.cloud_config_filename: 'cloud_api.conf'
   app.acli_config_filepath: "%app.data_dir%/%app.acli_config_filename%"
@@ -18,7 +17,6 @@ services:
       $acliConfigFilename: '%app.acli_config_filename%'
       # This should be root directory of the repository where acli is being invoked (not the root of acli itself).
       $repoRoot: "%app.repo_root%"
-      $sshDir: "%app.ssh_dir%"
       Webmozart\KeyValueStore\JsonFileStore $datastoreCloud: '@datastore.cloud'
       Webmozart\KeyValueStore\JsonFileStore $datastoreAcli: '@datastore.acli'
     public: true
--- a/src/Command/Api/ApiCommandHelper.php
+++ b/src/Command/Api/ApiCommandHelper.php
@@ -115,8 +115,7 @@ class ApiCommandHelper {
     string $repoRoot,
     ClientService $cloudApiClientService,
     LogstreamManager $logstreamManager,
-    SshHelper $sshHelper,
-    string $sshDir
+    SshHelper $sshHelper
   ) {
     $this->cloudConfigFilepath = $cloudConfigFilepath;
     $this->localMachineHelper = $localMachineHelper;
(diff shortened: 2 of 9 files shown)


danepowell:
I think this is the right approach but only the home directory should be mutable, everything else should be relative to that. Will need to adjust tests to accommodate that.

danepowell:
@imnetworku since you've worked with Symfony service containers before, does this approach make sense to you?

To recap: we store some config in the user's home directory, such as `~/.ssh`, `~/.acquia/cloud_api.conf`. We previously defined these paths via `config/services.yml` based on the value of the `$HOME` environment variable. The problem is that Windows doesn't always have HOME, so we need to set the HOME directory dynamically.

The problem is that you can't set a service parameter dynamically, and almost _all other parameters_ depend on it. So I think they can no longer be parameters at all. I was going to create methods on LocalMachineHelper to set each of them dynamically.

imnetworku:
@danepowell That is one possibility, but I think there could be a workaround to make it seem like the service parameters are set dynamically. Let me do a couple PoCs and get back to you
> @imnetworku since you've worked with Symfony service containers before, does this approach make sense to you?
> 
> To recap: we store some config in the user's home directory, such as `~/.ssh`, `~/.acquia/cloud_api.conf`. We previously defined these paths via `config/services.yml` based on the value of the `$HOME` environment variable. The problem is that Windows doesn't always have HOME, so we need to set the HOME directory dynamically.
> 
> The problem is that you can't set a service parameter dynamically, and almost _all other parameters_ depend on it. So I think they can no longer be parameters at all. I was going to create methods on LocalMachineHelper to set each variable (no longer parameters) dynamically. Then services will need to call `LocalMachineHelper::getSshDir()`, `LocalMachineHelper::getDataDir()`, and such as appropriate rather than receiving a parameter.

imnetworku:
Picking up for manual test

imnetworku:
Manual Test Failed.

In the left side is the failing test on Windows 10 without HOME environment variable defined. After the failure, tested again with HOME system environment variable created and it works.

image

imnetworku:
Manual Test Failed.

Without an explicitly defined HOME environment variable, the output is the same.
image

config/services.yml
```diff
@@ -3,7 +3,6 @@ parameters:
   env(ACLI_REPO_ROOT): "%kernel.project_dir%"
   app.data_dir: "%env(HOME)%/.acquia"
```
imnetworku, line 4: @danepowell This is the line that is causing the failure.

imnetworku:
@danepowell I went through the transitive changes needed to get it to work on windows... it works but i'm not sure how elegant you think it is... 
https://github.com/imnetworku/cli/commit/c0c065c728d111441005d80e588dea9445e2c389
image

danepowell:
Thanks, I'll check your feature branch. The alternative is that we call LocalMachineHelper statically to find and set the HOME environment variable (if it's not already set) in bin/acli, before the service container gets compiled.

danepowell:
I think there might still be value in what I attempted here (consolidating parameters and arguments into LocalMachineHelper), but it's not necessary to fix the issue at hand (Windows Powershell). See #187 for a much simpler approach.

PR #171, opened 2020-06-17
Move event listener to a service

grasmash:
I've tried creating both an event listener and subscriber, but I can't get any of the methods (apart from getSubscribedEvents) to actual fire.

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Acquia\Cli\EventListener;
+
+use Acquia\Cli\Exception\AcquiaCliException;
+use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\Console\Event\ConsoleErrorEvent;
+
+class ExceptionListener
+{
+
+  public function onConsoleError(ConsoleErrorEvent $event) {
+    $exitCode = $event->getExitCode();
+    $error = $event->getError();
+    // Make OAuth server errors more human-friendly.
+    if ($error instanceof IdentityProviderException && $error->getMessage() === 'invalid_client') {
+      $event->setError(new AcquiaCliException('Your Cloud API credentials are invalid. Run acli auth:login to reset them.',
+        [], $exitCode));
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Acquia\Cli\Exception\AcquiaCliException;
+use League\OAuth2\Client\Provider\Exception\IdentityProviderException;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\Console\Event\ConsoleErrorEvent;
+
+class ExceptionListener
+{
+
+  public function onConsoleError(ConsoleErrorEvent $event) {
```
grasmash, line 13: Tried this based on https://symfony.com/blog/new-in-symfony-4-4-simpler-event-listeners

src/EventSubscriber/ExceptionSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+use Symfony\Component\Console\Event\ConsoleEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class ExceptionSubscriber implements EventSubscriberInterface
+{
+
+  /**
+   * @return array
+   */
+  public static function getSubscribedEvents() {
```
grasmash, line 18: Based on https://symfony.com/doc/current/event_dispatcher.html#creating-an-event-subscriber

grasmash: This method gets called.

src/EventSubscriber/ExceptionSubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+      ConsoleEvents::COMMAND => [
+        ['processCommand', 10],
+      ],
+      ConsoleEvents::TERMINATE => [
+        ['processTerminate', 10],
+      ],
+    ];
+  }
+
+  public function processException(ConsoleErrorEvent $event) {
```
grasmash, line 32: This method does not.

src/EventListener/ExceptionListener.php
grasmash: This never seems to get called.

grasmash:
I'm giving up on this.

PR #178, opened 2020-06-19
Fixes #177: Allow [site] and [site].[env] to be used rather than appU…

### LINKED ISSUE #177: Allow [site] and [site].[env] to be used rather than appUuid or EnvUuid
We already do this for the remote:* commands: https://github.com/acquia/cli/blob/master/src/Command/Remote/SshBaseCommand.php#L47

We also already substitute applicationUuid when a repo is linked: https://github.com/acquia/cli/blob/master/src/Command/Api/ApiCommandBase.php#L57

So for instance, it'd be nice if both of these worked:
```
acli api:applications:find eemgrasmick
acli api:applications:find 257a5440-22c3-49d1-894d-29497a1cf3b8
```

Or:
```
acli api:environments:domain-clear-varnish 21351-257a5440-22c3-49d1-894d-29497a1cf3b8 eemgrasmickdev.prod.acquia-sites.com
acli api:environments:domain-clear-varnish eemgrasmick.dev eemgrasmickdev.prod.acquia-sites.com
```

And not just for api:* commands, for all commands.

grasmash:
…uid or EnvUuid.


grasmash:
This works, but no test coverage yet.

grasmash:
As a follow up, I would like to update the usage examples to show that these aliases can be used

PR #184, opened 2020-06-23
Hide update command in IDEs.

grasmash:


anavarre:
Yes, I debated whether to make this available in IDEs and after thinking about it it's best we control which specific version is installed, especially if we have a regression with a newer version.

PR #198, opened 2020-07-10
Use Acquia project templates for acli new

danepowell:
@anavarre @grasmash This means that we will be creating Drupal 9 projects by default (instead of Drupal 8). I think that's desirable, but it's a big change.

### DIFF
--- a/src/Command/NewCommand.php
+++ b/src/Command/NewCommand.php
@@ -32,8 +32,8 @@ class NewCommand extends CommandBase {
    * @throws \Exception
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
-    $this->output->writeln('Acquia recommends most customers use <comment>acquia/drupal-recommended-project</comment>, which includes useful utilities such as Acquia Connector.');
-    $this->output->writeln('<comment>acquia/drupal-minimal-project</comment> is the most minimal application that will run on Acquia Cloud.');
+    $this->output->writeln('Acquia recommends most customers use <options=bold>acquia/drupal-recommended-project</>, which includes useful utilities such as Acquia Connector.');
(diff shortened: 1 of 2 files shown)


anavarre:
Still, that's the right course of action IMO, so I'm +1 on this. Also, you put `ads new` here.

danepowell:
Lol, thanks... fixed the title

src/Command/NewCommand.php
```diff
@@ -32,23 +32,22 @@ protected function configure() {
    * @throws \Exception
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
+    $this->output->writeln('Acquia recommends most customers use <comment>acquia/drupal-recommended-project</comment>, which includes useful utilities such as Acquia Connector.');
```
grasmash, line 35: We've started to use `<options=bold></>` rather than `<comment>` for this.

tests/phpunit/src/Commands/NewCommandTest.php
```diff
@@ -94,11 +94,6 @@ public function testNewCommand($project): void {
     $this->assertStringContainsString($project, $output);
     $this->assertStringContainsString('New 💧Drupal project created in ' . $this->newProjectDir, $output);
 
-    if ($project === 'drupal/recommended-project') {
-      $composer_json = file_get_contents(Path::join($this->newProjectDir, 'composer.json'));
-      $this->assertStringNotContainsString('web/', $composer_json);
-      $this->assertStringContainsString('docroot/', $composer_json);
```
grasmash, line 100: I think we should still assert that we have the expected composer.json in the right place.

danepowell: I don't think there's any value in this test, because we've hardcoded a composer.json in the text fixture, so all we'd testing is the fixture, nothing related to the business logic in the code.

PR #200, opened 2020-07-17
Fixes #199: Create an ide:php command.

### LINKED ISSUE #199: Create an ide:php command
Currently, to switch PHP versions in IDEs, you need to do the following and restart PHP:

```
echo "7.4" >> /home/ide/configs/php/.version
```

To enable Xdebug, you need to uncomment a line under `/home/ide/configs/php/xdebug.ini` and restart PHP.

We should make all PHP operations easier with Acquia CLI, by introducing an `ide:php` command. The command should do the following:

* Allow switching PHP versions (7.2 to 7.4)
* Enable/disable Xdebug

Note: this command should probably only be visible from within an IDE.

grasmash:


src/Command/Ide/IdePhpVersionCommand.php
```diff
@@ -0,0 +1,52 @@
...
+
+    return 0;
+  }
+
+  /**
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function restartBash() {
+    $this->logger->notice('Restarting bash...');
+    passthru('exec bash -l');
```
grasmash, line 49: Note that exec(), shell_exec(), and Process() do not work for this.

joazvsoares: We suggested `bash` command for simplicity, but we only need to update environment variables to reflect the change when using PHP CLI:
```
ide:~/project$ cat ~/configs/php/.version 
7.3
ide:~/project$ php --version
PHP 7.3.18 (cli) (built: Jun  1 2020 19:39:17) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.18, Copyright (c) 1998-2018 Zend Technologies
ide:~/project$ echo "7.4" > ~/configs/php/.version
ide:~/project$ export PHP_VERSION=7.4
ide:~/project$ export PATH="/usr/local/php${PHP_VERSION}/bin:${PATH}"
ide:~/project$ php --version
PHP 7.4.6 (cli) (built: Jun  1 2020 17:56:08) ( NTS )
Copyright (c) The PHP Group
Zend Engine v3.4.0, Copyright (c) Zend Technologies
ide:~/project$

grasmash: Great, I would prefer a more lightweight solution. I’ll change this to set the environmental variable instead.

grasmash: @joazvsoares Unfortunately, the env var approach doesn't seem to work. The `putenv()` changes are not reflected in the terminal. I also don't see a `PHP_VERSION` variable defined at all with `printenv | grep PHP` in the IDE.

grasmash: Hm yeah the `putenv()` docs read: 
> Adds setting to the server environment. The environment variable will only exist for the duration of the current request. At the end of the request the environment is restored to its original state.

grasmash: So we're back to restarting bash.

grasmash: @anavarre The only downside to this that I can see is that restarting bash wipes out your terminal history, so you can't "scroll up" to previous commands after running it.

anavarre: There doesn't need to be any. You could copy `/home/ide/.bash_history`, run the command, and write it back.

danepowell: We ran into this in the early days of BLT, there's no way for PHP to modify env vars outside of its current process.

joazvsoares: On second thought, even if you could update the environment variable, the change would not apply for other terminal tabs that the user might have already opened, the same happens to the `bash` command.

The ideal alternative would be to wrap our PHP executable into a shim script in Cloud IDEs, the script will read the content of the `.version` file and run the PHP version accordingly (https://github.com/sptndc/phpenv#understanding-shims), no need to update the environment variable or run `bash` command with this implementation.

It's worth mentioning that this only happens for the PHP CLI because we're trying to reflect the change in the current terminal session in order to offer a seamless experience without asking the user to open a new terminal tab.

PR #210, opened 2020-07-24
Fixes DT-1396: Refresh should match PHP version in IDE.

grasmash:


tests/phpunit/src/Commands/Ide/Wizard/IdeWizardDeleteSshKeyCommandTest.php
```diff
@@ -22,15 +22,18 @@ class IdeWizardDeleteSshKeyCommandTest extends IdeWizardTestBase {
...
+    $this->clientProphecy->request('get', '/ides/' . $this::$remote_ide_uuid)->willReturn($ide_response)->shouldBeCalled();
     $ide = new IdeResponse((object) $ide_response);
     $mock_body = $this->mockListSshKeysRequestWithIdeKey($ide);
 
     $this->mockGetIdeSshKeyRequest($ide);
     $this->mockDeleteSshKeyRequest($mock_body->{'_embedded'}->items[0]->uuid);
 
     // Create the file so it can be deleted.
-    $ssh_key_filename = $this->command->getSshKeyFilename($this->remote_ide_uuid);
+    $ssh_key_filename = $this->command->getSshKeyFilename($this::$remote_ide_uuid);
```
danepowell, line 36: I don't understand how this line of code works.

`$this` is an object context, but `::` is a scope operator and typically used in a static context (i.e. `self::whatever`). I guess maybe `$this::whatever` works, but it doesn't seem right.

But then where is `$remote_id_uuid` defined? It doesn't appear to be global or defined anywhere in this method.

grasmash: Based on PHP documentation, this is the correct way to reference a statically defined object property. Unlike self, this allows us to access the member property whether it’s defined on the current class or one of the parent classes.

PR #213, opened 2020-07-29
Update README.md

alex-moreno:
Updated with latest release. Maybe would be better to point to /latest, but could not find for some reason the right link.

Also reduced in one line the steps, but that's maybe a bit pedantic from me

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,13 @@ Download the latest `acli.phar` file from the [releases](https://github.com/acqu
 
 For example, you can run following simple script:
 ```bash
-curl -OL https://github.com/acquia/cli/releases/download/v1.0.0-rc3/acli.phar
-mv acli.phar acli
-chmod +x acli
+curl -OL https://github.com/acquia/cli/releases/download/v1.0.0-rc4/acli.phar
+chmod +x acli.phar
 ```
 
 Next, place it somewhere globally accessible on your machine. For instance:
 ```
-mv acli /usr/local/bin/
+mv acli.phar /usr/local/bin/acli
 ```
 
 Acquia CLI cannot and should not be installed via Composer. It is distributed only as a [self-contained Phar archive](https://www.php.net/manual/en/phar.using.intro.php) in order to avoid dependency conflicts.


README.md
```diff
@@ -13,14 +13,13 @@ Download the latest `acli.phar` file from the [releases](https://github.com/acqu
...
-mv acli.phar acli
-chmod +x acli
+curl -OL https://github.com/acquia/cli/releases/download/v1.0.0-rc4/acli.phar
+chmod +x acli.phar
 ```
 
 Next, place it somewhere globally accessible on your machine. For instance:
 ```
-mv acli /usr/local/bin/
+mv acli.phar /usr/local/bin/acli
```
alex-moreno, line 22: reduced in one line the steps, but that's maybe a bit pedantic from me

README.md
```diff
@@ -13,14 +13,13 @@ Download the latest `acli.phar` file from the [releases](https://github.com/acqu
 
 For example, you can run following simple script:
 ```bash
-curl -OL https://github.com/acquia/cli/releases/download/v1.0.0-rc3/acli.phar
-mv acli.phar acli
-chmod +x acli
+curl -OL https://github.com/acquia/cli/releases/download/v1.0.0-rc4/acli.phar
```
alex-moreno, line 16: Maybe would be better to point to /latest, but could not find for some reason the right link

anavarre: The problem is not linking to the latest release, it's finding how to [link to the correct PHAR file](https://gist.github.com/steinwaywhw/a4cd19cda655b8249d908261a62687f8). And that requires a pretty advanced one-liner so I'm not sure we can afford doing that for newcomers to Acquia CLI.

Perhaps that should be part of the release runbook to modify the README file.

grasmash: I’m pretty sure that the link for the latest asset doesn’t work because we don’t have a stable release yet. Once we have a stable release, it should start working and we can update this file. Until then, let’s just incrementally update the file manually.

README.md
grasmash: Why remove the permissions change?

alex-moreno: it's still there, line 17 (chmod +x acli.phar), but you are saving one line of code doing the move from acli.phar to acli when moving it to the final place.

Again, a bit pedantic, so I can leave it as it was before

README.md
alex-moreno: that thought of no stable release came to my mind

PR #233, opened 2020-08-17
Remove custom code from bin/acli

danepowell:
Content in this file can't be tested, so it should be as close to stock as possible. Stuff like the progress bar format, repo root detection, and exception handling should be moved to dedicated services or static methods that can be tested.

Matt already took care of the event dispatcher in https://github.com/acquia/cli/pull/171

Follow-up from https://github.com/acquia/cli/pull/118


danepowell:
Hmm, I'm getting bogged down trying to figure out how to mock the project root detection for tests

danepowell:
Ready for review now.

Issue #228, opened 2020-08-14
PHP Notice:  Trying to get property 'sshUrl' of non-object in phar:///usr/local/bin/acli/src/Helpers/SshHelper.php on line 153


anavarre:
Ran into the below issue when testing RC6 on an ACSF test subscription. It also happens with RC1 so IDK under which conditions the error shows up because in non-ACE IDEs I have it working well.
```
ide:~$ ./acli.phar ide:wizard:ssh-key:create-upload
    ✔ Creating a local SSH key
    ✔ Uploading the local key to Acquia Cloud
    ✔ Adding the SSH key to local keychain
    ⠹ Waiting for the key to become available on Acquia Cloud web servers
PHP Notice:  Trying to get property 'sshUrl' of non-object in phar:///home/ide/acli.phar/src/Helpers/SshHelper.php on line 153
    ⣰ Waiting for the key to become available on Acquia Cloud web servers
PHP Notice:  Trying to get property 'sshUrl' of non-object in phar:///home/ide/acli.phar/src/Helpers/SshHelper.php on line 153
    ⡇ Waiting for the key to become available on Acquia Cloud web servers
(snip)
```

The key looks to be uploaded correctly though.

danepowell:
I just reproduced this on an ACSF subscription.

anavarre:
I tried the Acquia CLI build after #237 was merged but unfortunately I'm still seeing the issue.

PR #237, opened 2020-08-19
Fix #227: Errors refreshing ACSF site

### LINKED ISSUE #227: Issues refreshing ACSF sites
ide-acsf

@danepowell this one looks critical as it's blocking an ACSF customer from pulling the site locally.

danepowell:
Fixes #227  
--------

Motivation
----------
#227 describes an error when running `acli refresh` for ACSF sites. I believe this to be a regression caused by https://github.com/acquia/cli/pull/74 . This caused SSH command output to be printed to the screen, which apparently also prevents `$process->getOutput()` from capturing it.

Proposed changes
---------
Suppress SSH command output, so it can be captured programmatically. This is ideal for the human (who doesn't need to see a bunch of gobbledygook), and also fixes the error.

Alternatives considered
---------
- Change the default for [SshHelper::executeCommand()](https://github.com/acquia/cli/blob/v1.0.0-rc6/src/Helpers/SshHelper.php#L40) to _not_ print output. @joazvsoares @imnetworku do you have an opinion on this? Most places in the code that call `SshHelper::executeCommand()` set print_output to FALSE, should we make FALSE the default?
- Change the SSH process to both print and capture output. Might be useful at some point, but not right now.

Testing steps
---------
Run `acli refresh` on an ACSF application and verify that the error does not recur.

### DIFF
--- a/src/Command/RefreshCommand.php
+++ b/src/Command/RefreshCommand.php
@@ -572,7 +572,7 @@ class RefreshCommand extends CommandBase {
   protected function getAcsfSites($cloud_environment): array {
     $sitegroup = self::getSiteGroupFromSshUrl($cloud_environment);
     $command = ['cat', "/var/www/site-php/$sitegroup.{$cloud_environment->name}/multisite-config.json"];
-    $process = $this->sshHelper->executeCommand($cloud_environment, $command);
+    $process = $this->sshHelper->executeCommand($cloud_environment, $command, FALSE);
     if ($process->isSuccessful()) {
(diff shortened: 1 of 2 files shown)


imnetworku (approved):

anavarre (approved):
If it also worked for Joaz as part of the AMA tests then it gives me high confidence this is our fix.

@danepowell isn't this also fixing #228?

danepowell:
I suspect it might also fix #228 but I haven't specifically tested for it.

PR #239, opened 2020-08-25
Fix #228: PHP notice creating SSH key on ACSF

### LINKED ISSUE #228: PHP Notice:  Trying to get property 'sshUrl' of non-object in phar:///usr/local/bin/acli/src/Helpers/SshHelper.php on line 153

danepowell:
Fixes #228  
--------

Motivation
----------
See #228 

Proposed changes
---------
Use the environment detector to detect a dev environment rather than looking for a hardcoded string ('dev').

Alternatives considered
---------
Copying the regex or hardcoded dev environment names from the environment detector... silly :smile: 

Testing steps
---------
Apply this PR as a patch, run `ide:wizard:ssh-key:create-upload` against an ACSF subscription, verify that notices are gone

@anavarre can you please test and verify this fixes the issue for you?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
   "require": {
     "php": "^7.3",
     "ext-json": "*",
-    "acquia/drupal-environment-detector": "^1.1.1",
+    "acquia/drupal-environment-detector": "^1.2.0",
     "doctrine/cache": "^1.10",
     "drupol/phposinfo": "^1.6.5",
     "padraic/phar-updater": "^1.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,26 +4,31 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "fa7ad899b4c9a8f90310cd202b3f6617",
+    "content-hash": "388c3b763ab6a6236b4fd634b7eab242",
     "packages": [
         {
             "name": "acquia/drupal-environment-detector",
-            "version": "v1.1.1",
+            "version": "v1.2.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/acquia/drupal-environment-detector.git",
-                "reference": "d2a871f61a3c575e11063317f03c7d13e0ec23f0"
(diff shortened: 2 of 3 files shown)


anavarre:
I created a new IDE specifically to test this PR but unfortunately I ran into the same exact issue and even ended up running into a timeout (which I know is unrelated)

```
    ⡇ Waiting for the key to become available on Acquia Cloud web servers
PHP Notice:  Trying to get property 'sshUrl' of non-object in phar:///usr/local/bin/acli/src/Helpers/SshHelper.php on line 153
    ✔ Waiting for the key to become available on Acquia Cloud web servers
Timed out after 15 minutes!
```

anavarre (changes requested):
Per my comment above.

anavarre:
Any other idea how we could be solving this @danepowell ? I'd like to ship a new Acquia CLI version as soon as we can so the experience with ACSF is restored to normal operations.

danepowell:
@anavarre did you make sure to run `composer install` after checking out this feature branch? I assume you were testing by checking out the ACLI feature branch in the IDE. It has new composer dependencies, if you don't `composer install` you'll get the old errors.

I just tested on an ACSF sub again and got no notices.

anavarre:
After testing this more I cannot reproduce the issue. Let's assume I did something wrong the first time when I tested the fix. Looks good, ship it :smile:

PR #249, opened 2020-09-09
Fix #247: Rename api:environments:update --version parameter

### LINKED ISSUE #247: api:environments:update --version is twice defined
The help section of api:environments:update seems to have two `version` parameters. One is the default to show the `acli` version and the other is to attempt to set the php or node version per the internal documentation.

```
api:environments:update [--version VERSION]
```

danepowell:
Fixes #247 
--------

Motivation
----------
See #247. Running `acli help api:environments:update` results in an error because version is defined both via the CLI and API.

Proposed changes
---------
This renames the API `--version` to `--lang_version`.

Alternatives considered
---------
If this becomes a pattern with other parameters, we might want a full translation layer that renames conflicting parameters either automatically (if it detects a conflict) or statically (based on hardcoded mappings). Or at least use a constant utility class between the API command base and helper.

Testing steps
---------
- Run `acli cache:clear`
- Run `acli help api:environments:update`.
@eporama if you're comfortable running dev builds, you're welcome to test.

### DIFF
--- a/src/Command/Api/ApiCommandBase.php
+++ b/src/Command/Api/ApiCommandBase.php
@@ -7,14 +7,7 @@ use Acquia\Cli\Exception\AcquiaCliException;
 use AcquiaCloudApi\Exception\ApiErrorException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Validator\Constraints\Length;
-use Symfony\Component\Validator\Constraints\NotBlank;
-use Symfony\Component\Validator\Constraints\Regex;
-use Symfony\Component\Validator\Constraints\Uuid;
-use Symfony\Component\Validator\Constraints\UuidValidator;
-use Symfony\Component\Validator\Constraints\Valid;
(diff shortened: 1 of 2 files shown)


anavarre:
Maybe I'm not testing this correctly but after checking out the branch locally and running `composer install` I'm still seeing the same issue

```
acli api:environments:update SITEGROUP.ENV --memory_limit=128

                                             
  An option named "version" already exists.  
                                             

api:environments:update [--version VERSION] [--memory_limit MEMORY_LIMIT] [--opcache OPCACHE] [--apc APC] [--interned_strings_buffer INTERNED_STRINGS_BUFFER] [--max_execution_time MAX_EXECUTION_TIME] [--max_post_size MAX_POST_SIZE] [--max_input_vars MAX_INPUT_VARS] [--sendmail_path SENDMAIL_PATH] [--varnish_over_ssl VARNISH_OVER_SSL] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [--] <environmentId>
```

danepowell:
Yeah we'll have to fix that error separately, so that's a success :)

PR #252, opened 2020-09-15
Fixes #248: api:environments:update is not submitting integers as integers.

### LINKED ISSUE #248: api:environments:update is not submitting integers as integers
```
$ acli api:environments:update eeepeterson.dev --memory_limit=192
{
    "error": "validation_failed",
    "message": {
        "memory_limit": "192 is not a valid memory limit. It should be an integer."
    }
}
```

grasmash:


grasmash:
I can verify that this does correctly set memory_limit as an integer before passing it to Guzzle. However, it does not seem to resolve the issue, which makes me wonder if there could be an issue in Cloud API.

grasmash:
This could be an upstream issue in https://github.com/typhonius/acquia-php-sdk-v2.

This fails with the same error:

```
$env = new Environments($acquia_cloud_client);
$env->update('redacted', ['memory_limit' => 192]);
```

grasmash:
https://github.com/typhonius/acquia-php-sdk-v2/issues/85

PR #284, opened 2020-10-01
Notify users of available update.

grasmash:


anavarre:
Can we cache this so we don't always make an outbound request to return this info?

src/Helpers/UpdateHelper.php
```diff
@@ -0,0 +1,84 @@
...
+    $this->client = $client;
+  }
+
+  /**
+   * @return \GuzzleHttp\Client
+   */
+  public function getClient(): Client {
+    if (!isset($this->client)) {
+      $stack = HandlerStack::create();
+      $stack->push(new CacheMiddleware(), 'cache');
```
grasmash, line 39: @anavarre Adding caching here. It should respect GitHub's cache headers. Will need to validate that it's actually limiting outgoing requests.

anavarre: Thanks, I'm watching this PR. When you think it's ready, let me know how I can test it, maybe?

.travis.yml
```diff
@@ -33,6 +33,7 @@ jobs:
         - cmd.exe //c "RefreshEnv.cmd"
         # Emulate RefreshEnv to pick up new bin paths in Git Bash (the native Travis CI shell).
         - eval $(powershell -NonInteractive -Command 'write("export PATH=`"" + ([Environment]::GetEnvironmentVariable("PATH","Machine") + ";" + [Environment]::GetEnvironmentVariable("PATH","User")).replace("\","/").replace("C:","/c").replace(";",":") + ":`$PATH`"")')
+        - composer selfupdate --preview
```
grasmash, line 36: Note that this uses Composer 2.

danepowell (changes requested):
I'm impressed with the amount of work you've put into this in such a short time! :smile: 

I have two concerns though:

It seems like we can only build ACLI using Composer 2 now. When I run `composer install` with Composer 1 I get a dependency conflict in the typhonius packages. That's going to be a huge pain because most of our other stuff works the other way (i.e. with Composer 1 but not Composer 2)

Additionally, this adds 200-500ms to every command in my testing. I'm assuming this is because network calls aren't getting cached. Just conceptually, I don't see how they _could_ be cached because the phar is not writeable. The only way to cache them would be via a filestore (e.g. `~/.acquia/acquia-cli.json`)

grasmash:
We can definitely use a filestore for caching. I'm in the midst of local testing the network requests.

grasmash:
And yeah... I'm not sure why this fails with Composer 1.

grasmash:
Last code push makes all requests after the initial request ~300ms.

grasmash:
Using Composer 1.

danepowell (approved):

PR #290, opened 2020-10-06
Fixes #289: acli ide:service-restart should output something.

### LINKED ISSUE #289: acli ide:service-restart should output something
Currently, `acli ide:service-restart <service>` doesn't return anything even if it works fine. To know what's going on, users have to run the command in verbose mode, which is not intuitive.

```
$ ./bin/acli ide:service-restart apache2 --verbose
[notice] Restarting apache2...
[notice] Restarted apache2
```
The regular output should return the status of the service instead of only when in verbose mode.

grasmash:


grasmash:
Covered by #291

PR #293, opened 2020-10-07
Accept application and environment aliases.

grasmash:
Functional changes:

* log:tail now accepts aliases like `@eemgrasmick.dev`
* link now accepts aliases like `@eemgrasmick.dev`
* ide:* commands now accept aliases with and without `@` prefix
* remote:aliases:list will now accept an alias for the applicationUuid argument


anavarre:
Works great!

PR #294, opened 2020-10-08
Support large databases for refresh command.

grasmash:


grasmash:
@anavarre Can you test this? If it doesn't work, I'll try scp.

anavarre:
Unfortunately I ran into the same issue.

```
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 1519716680 bytes) in phar:///tmp/acli/vendor/symfony/process/Process.php on line 566
```

wimleers:
This explains https://acquia.slack.com/archives/G6ZCYUKUP/p1602164674042500 😂 

Thanks, @anavarre  for the ping, resulted in https://acquia.slack.com/archives/G812NSNP9/p1602171965006500 🥳

anavarre:
Tried the latest changes in this PR after I manually expanded my test DB size to ~5GB.

```
$ /tmp/cli/bin/acli refresh --no-code --no-scripts --no-files -vvv
[debug] Loading local project information...
[debug] Searching local datastore for matching project...
[debug] Matching local project found.
[debug] Cloud application UUID: <REDACTED_UUID>
[notice] A newer version of Acquia CLI is available. Run acli self-update to update to 1.0.0
Using Cloud Application <REDACTED_SITEGROUP>
Choose a Cloud Platform environment to copy from:
  [0] Dev (vcs: pipelines-build-master)
  [1] Stage (vcs: tags/2020-09-28)
 > 1
[debug] Using environment Stage <REDACTED_UUID>
```

After 1 hour, it seems I'm still blocked on importing the DB and there's no indication (even with `-vvv`) if something is happening or not. Seems like it should be way way faster if it worked.

grasmash:
So, I think we should use this snippet to fill the db: 
`drush ev "while(true){\Drupal::logger('system')->debug(\Drupal\Component\Utility\Random::string(64));};"`

My theory is that our former command, which repeated a single character rather than creating a random string, greatly reduces the performance of `mysqldump` and `gzip`. 

Your output suggests that it never actually finishes dumping the db. It never starts downloading it or importing it.

grasmash:
I've added a commit to show progress better. It's a bit messy, but we can clean it up once it works.

danepowell (approved):

tests/phpunit/src/Commands/RefreshCommandTest.php
```diff
@@ -585,8 +576,9 @@ protected function mockExecuteSshMySqlDump(
     $process->getOutput()->willReturn('dbdumpcontents');
     $ssh_helper->executeCommand(
         new EnvironmentResponse($environments_response),
-        ['MYSQL_PWD=heWbRncbAfJk6Nx mysqldump --host=fsdb-74.enterprise-g1.hosting.acquia.com --user=s164 profserv201dev | gzip -9'],
-        FALSE, 60 * 60
+        ['MYSQL_PWD=heWbRncbAfJk6Nx mysqldump --host=fsdb-74.enterprise-g1.hosting.acquia.com --user=s164 profserv201dev | pv --rate --bytes | gzip -9 > /mnt/tmp/profserv201dev/acli-mysql-dump-dev-profserv201dev.sql.gz'],
```
anavarre, line 579: Aren't we introducing a security issue by having real creds exposed?

anavarre:
Confirmed that refreshing a 4.3GB worked just fine on ACE. The progress bar and added verbosity is incredibly helpful. Filed a follow-up for ODEs - #306

Issue #296, opened 2020-10-08
Cache mapping of aliases -> uuids


anavarre:
E.g.

```
$ time acli api:environments:database-list <SITEGROUP>.<ENV>
0m8.907s
```
Using an alias instead of a UUID makes an extra API call and it'll inherently be slow.

We could add some caching so we have a mapping of aliases -> UUID that would prevent from querying the API always. This never changes so it's safe to cache.

danepowell:
Specifically, we should map both sitegroups to application IDs, and environments to environment IDs.

anavarre:
Fix in #298

PR #298, opened 2020-10-08
Cache command aliases.

grasmash:
Fix #298

### DIFF
--- a/bin/acli
+++ b/bin/acli
@@ -62,14 +62,13 @@ if (!getenv('HOME')) {
 
 $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
 
-// Handle a cache:clear pseudo command. This isn't implemented as a true console
+// Handle a clear-kerner-cache pseudo command. This isn't implemented as a true console
 // command because a stale or corrupted cache would render it unusable--
 // precisely when it is needed.
-if (in_array($input->getFirstArgument(), ['clear-internal-cache', 'cci'])) {
+if (in_array($input->getFirstArgument(), ['clear-kernel-cache', 'ckc'])) {
   $filesystem = new Filesystem();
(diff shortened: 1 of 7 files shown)


bin/acli
```diff
@@ -65,7 +65,7 @@ $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
 // Handle a cache:clear pseudo command. This isn't implemented as a true console
 // command because a stale or corrupted cache would render it unusable--
 // precisely when it is needed.
-if (in_array($input->getFirstArgument(), ['cache:clear', 'cc'])) {
+if (in_array($input->getFirstArgument(), ['clear-internal-cache', 'cci'])) {
   $filesystem = new Filesystem();
   $kernel_cache_dir = $kernel->getCacheDir();
   $system_cache_dir = Path::join(sys_get_temp_dir(), 'symphony-cache');
```
danepowell, line 71: I think we'd want to clear the "system cache" via the new clear-caches command. It's in a temporary system directory so it should be good/safe to clear.

I think we should rename this pseudo-command to `clear-kernel-cache` or something, and have it only clear the kernel cache (that lives in `app-dir/var`).

There's also a comment above this conditional that should be renamed.

danepowell (approved):

bin/acli
```diff
@@ -62,14 +62,13 @@ if (!getenv('HOME')) {
 
 $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
 
-// Handle a cache:clear pseudo command. This isn't implemented as a true console
+// Handle a clear-kerner-cache pseudo command. This isn't implemented as a true console
```
itafroma, line 65:
```suggestion
// Handle a clear-kernel-cache pseudo command. This isn't implemented as a true console
```

anavarre:
Fix #296

PR #299, opened 2020-10-08
Update README.md

grasmash:


README.md
```diff
@@ -10,7 +10,6 @@ Acquia CLI does not provide or manage local development environments. If you are
 ## Similar tools
 Several tools compliment or duplicate Acquia CLI functionality. Acquia CLI can safely be used with any of these tools, although some functionality may be duplicated.
 - [Acquia BLT](https://github.com/acquia/blt): Provides an automation framework for setting up and managing Drupal applications. Acquia BLT is installed in a Drupal application and provides support for just that application, while Acquia CLI should be installed globally and allows you to interact with any Acquia service. Acquia CLI is not an automation framework like Acquia BLT.
```
anavarre, line 12: Shouldn't we also remove BLT here?

grasmash: Why? BLT is not EOL.

anavarre: It's not this. Having BLT here makes it seem like it's a CLI tool. But is it, really? IMHO it's not because we have a CLI tool to manage it that it is related to Acquia CLI in any way.

grasmash: I don't really understand you but I'm going to merge this. We can remove BLT in another PR still.

PR #301, opened 2020-10-09
Don't link directories that aren't repos

danepowell:
Motivation
----------
I find it annoying when I run `acli` outside of a repository (for instance in my home directory) and it needlessly prompts me to link the current "repository" to a Cloud app. The CLI already knows if the current directory is in a repository, let's use that information to hide the prompt if it's irrelevant.

Proposed changes
---------
Don't prompt to link a directory if it's not a repository.


anavarre:
Yes, please!

danepowell:
This was closed without merging but I think Matt committed it elsewhere as part of a rebase

Issue #201, opened 2020-07-21
Create a deploy command


anavarre:
In Acquia CLI, there's no equivalent to `ads deploy`. This means in Cloud IDE users might be confused as to how they can deploy to any Cloud environment.

Here's how `ads deploy` currently works:

```
$ ads deploy
Select corresponding Acquia Cloud Application: mydrupalwebsite
? Select source:  Stage
? Select target environment:  Prod
? Are you sure you want to deploy to Prod? Yes
? Do you want to back up the databases before deploying the code? No
? Do you want to clear Varnish cache after the deploy? Yes

Summary:

 ›   Warning: The databases from Prod will not be backed up before deployment.
The environment Prod will be switched to progress.
Varnish cache will be cleared from Prod after deployment.

? Continue? Yes
  ✔ mydrupalwebsite: deploying progress to Prod
  ✔ Clearing Varnish cache

Successfully deployed progress to Prod
? Open the website? No
```

Perhaps initially in an IDE there could be a wrapper to `ads deploy` until we have a full-blown command.

grasmash:
I’m not convinced that this is valuable, and I think that it may be outside of a defined scope for this tool.

What are we trying to accomplish? I would expect that deployment would involve the following steps:

* git add, commit, and push
* check out tag or branch on a given cloud environment
* run post deploy commands like cache clear or data as updates

I’m not sure that any of those things make sense as the responsibility of Acquia CLI.

anavarre:
Real-life use case from an Acquia CLI user in IDEs:

- Pulled code, DB, files in the IDE
- Replaced codebase by fresh new install
- User blocked on uploading DB and files back to the Cloud Platform
- Got back to Dev Studio CLI to be unblocked

grasmash:
OK, it sounds like you’re envisioning a different scope for this command. Are you suggesting that the command should push files and database upstream? Can you write the user story and requirements?

aadityajshah:
I think `ads deploy` or a similar feature is definitely needed. For example, my current `dev` environment is running an older version of Drupal 8. In my CloudIDE I have created a new multi-site running Drupal 9. I should be able to push out the code to the repo, as well as update the DB as needed in this instance. Also, it should work without BLT as well 

I agree that it should not necessarily be an ACLI command, but it should allow for a `deploy` of some sort to be implemented through the CloudIDE.

grasmash:
For the sake of clarity, I think it might be better to do something like this:

[...]

PR #303, opened 2020-10-10
Rename refresh command to pull. Create pull sub command stubs.

grasmash:
No functional changes at all, just a rename of `refresh` to `pull:all` (with aliases to `refresh` and `pull`) and a few new non-functional stubs.

This does prep work for #201, which will introduce `push:*` commands.

image

### DIFF
--- a/src/Command/Pull/PullCodeCommand.php
+++ b/src/Command/Pull/PullCodeCommand.php
@@ -41,4 +41,5 @@ class PullCodeCommand extends PullCommandBase {
   protected function execute(InputInterface $input, OutputInterface $output) {
 
   }
+
 }
--- a/src/Command/Pull/PullCodeCommand.php
+++ b/src/Command/Pull/PullCodeCommand.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Acquia\Cli\Command\Pull;
+
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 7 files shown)


grasmash:
We can change the stub commands to be hidden for the time being.

anavarre:
Can you group arguments to mimic e.g. `acli refresh --no-scripts --no-code --no-files`?

grasmash:
Sorry, I don’t think I understand what you mean.

grasmash:
Also, I’m not sure that we should really have those arguments anymore. Now that we have dedicated commands for each of the subsets, it seems like you should just run those if you don’t want everything.

anavarre:
> Now that we have dedicated commands for each of the subsets, it seems like you should just run those if you don’t want everything.

That's my point. For convenience you want one command to do everything in one go (that's what arguments are meant for). If you need to run 3 commands instead of one, then we've not improved your quality of life, which means we lost something in the process.

I suggested before we should have this both ways: only **refresh** X, Y and Z, or **exclude** X, Y and Z to have everything covered.

anavarre:
Here's an example of what I have in mind: `acli pull --except=database,files`: Will only pull your codebase and run `composer install`

I'm also concerned that if we don't exclude the DB or files easily, `acli` will just ahead and try to refresh everything again even if it's not needed.

grasmash:
Sure. I'd say merge this first, then let's change the command. This PR is intentionally scoped to exclude all functional changes.

PR #305, opened 2020-10-12
Split pull:all functionality into sub commands, add push:* commands.

grasmash:


src/Command/Push/PushCodeCommand.php
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * @param \Symfony\Component\Console\Input\InputInterface $input
+   * @param \Symfony\Component\Console\Output\OutputInterface $output
+   *
+   * @return int 0 if everything went fine, or an exit code
+   * @throws \Exception
+   */
+  protected function execute(InputInterface $input, OutputInterface $output) {
+    $output->writeln("Please use <options=bold>git</> to push code changes upstream.");
```
danepowell, line 36: Is this just a placeholder? I think it would be better to have no `push:code` command, rather than a command that does nothing and exits.

danepowell (approved):

anavarre:
FTR I can't test this PR until #307 is fixed, which will then allow me to confirm the fix for #306

PR #308, opened 2020-10-14
Composer update

danepowell:
This adds Composer 2 compatibility by updating some upstream packages that recently added Composer 2 compatibility.

This only affects users developing on Acquia CLI, it does not affect end users.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "symfony/validator": "^5.1",
     "symfony/yaml": "^5.1@dev",
     "typhonius/acquia-logstream": "dev-master",
-    "typhonius/acquia-php-sdk-v2": "dev-master as 2.0.13",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -36,7 +36,7 @@
     "symfony/validator": "^5.1",
     "symfony/yaml": "^5.1@dev",
     "typhonius/acquia-logstream": "dev-master",
-    "typhonius/acquia-php-sdk-v2": "dev-master as 2.0.13",
+    "typhonius/acquia-php-sdk-v2": "^2.0.14",
```
danepowell, line 39: Changing to a stable release was necessary to fix a Travis CI error (not clear on the precise cause). And anyway, stable versions are better, and Adam is pretty responsive about cutting stable releases.

PR #309, opened 2020-10-14
DX-2533: Added confirmation question prompt and default responses

danepowell:
Motivation
----------
We received product feedback from @ba66e77 that the confirmation questions asked by Acquia CLI are confusing because they don't tell you what the available/default responses are (y/n/yes/no/whatever). For the record, Symfony basically interprets anything starting with a "y" as "yes", but being more explicit can't hurt.

Proposed changes
---------
Switch from the Symfony question helper to the higher-level style component, which automatically formats confirmation questions to include suggested and default responses.

This has the effect of changing how questions are formatted.

Before:
before

After:
after

Alternatives considered
---------
Keep using the lower-level question helpers and manually modify each invocation to add y/n and defaults.

Testing steps
---------
Run any command that generates a confirmation question, such as `acli auth:login` when you haven't set telemetry preferences.

@grasmash @anavarre how do you feel about this change, before I go through and convert all the remaining cases, as well as the failing tests?

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -66,6 +66,11 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
   protected $output;
 
   /**
+   * @var \Symfony\Component\Console\Style\SymfonyStyle
+   */
+  protected $io;
+
+  /**
    * @var \Symfony\Component\Console\Helper\FormatterHelper*/
   protected $formatter;
 
@@ -214,6 +219,7 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
   protected function initialize(InputInterface $input, OutputInterface $output) {
     $this->input = $input;
     $this->output = $output;
+    $this->io = new SymfonyStyle($input, $output);
     // Register custom progress bar format.
     ProgressBar::setFormatDefinition(
       'message',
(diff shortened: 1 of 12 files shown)


anavarre:
This change works for me. I'm all for improving the UX.

src/Command/CommandBase.php
```diff
@@ -261,8 +261,8 @@ public function checkAndPromptTelemetryPreference(): void {
     if (!isset($send_telemetry) && $this->input->isInteractive()) {
       $this->output->writeln('We strive to give you the best tools for development.');
       $this->output->writeln('You can really help us improve by sharing anonymous performance and usage data.');
-      $question = new ConfirmationQuestion('<question>Would you like to share anonymous performance usage and data?</question> ', TRUE);
-      $pref = $this->questionHelper->ask($this->input, $this->output, $question);
+      $style = new SymfonyStyle($this->input, $this->output);
```
grasmash, line 264: This looks good, but I'm just confused by Symfony here. If this is the right way to do things it seems like we should change our approach to output in a few places.

Looking at the [Symfony examples](https://symfony.com/doc/current/console/style.html), it seems like we should be naming the `$style` variable `$io` here:

```
        $io = new SymfonyStyle($input, $output);
        $io->title('Lorem Ipsum Dolor Sit Amet');
```

grasmash: Yeah I looked into it more and it does seem like we should audit our code to see where else we can use SymfonyStyle in order to reduce code repetition and improve usability.

danepowell: I chose `$style` based on how [Symfony itself uses the style component](https://github.com/symfony/console/blob/3b597067ee2b07ef6fbbebfe2afa0e88fae9a8a7/Application.php#L268), but I'm not picky on names.

danepowell:
I converted all of our questions to use the Symfony Style (except a couple that used lower-level options and normalizers not supported by Symfony Style).

We could probably convert all of our `$output->writeln()` calls to use `$io->text()`, `$io->success()`, etc... but that should happen in a separate ticket.

PR #310, opened 2020-10-14
DX-1605: Add usage examples

danepowell:
Motivation
----------
Most commands have no usage examples except for those automatically generated by Symfony for each argument. It's not clear to users, for instance, that they can use either an application UUID or sitegroup alias for any command with an app UUID argument.

Proposed changes
---------
Instead of manually adding app / environment ID arguments and usage examples to every command individually, encapsulate these in a common method. See before and after examples for the two commands refactored so far:
link
logtail

Alternatives considered
---------
Write the usage example for each command individually. That's a lot of unnecessary repetition.

Testing steps
---------
Run `acli help link` or `acli help log:tail`.

@grasmash @anavarre are you okay with this general approach? If so I'll finish refactoring the rest of the commands.

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -28,6 +28,7 @@ use Symfony\Component\Cache\Adapter\PhpArrayAdapter;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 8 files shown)


anavarre:
Much better. I'm all-in!

grasmash:
Yes this looks good to me. It seems that we should do it for both application and environment uuids.

danepowell:
I don't understand why coverage went down on this, I think Coveralls is just flipping out. It says the delta is in a Pull command, but I'm not even touching a pull command in this PR...

PR #321, opened 2020-10-19
Fixes 297: Missing access token causes lengthy API calls.

grasmash:


grasmash:
Looks like this broke Windows:
https://github.com/typhonius/acquia-php-sdk-v2/commit/a301ebd5ef18d7cd2f314dfca1fda9e660bb6155

With:
```
Error: Call to undefined function AcquiaCloudApi\Connector\posix_getpwuid()
```

grasmash:
Blocked by https://github.com/typhonius/acquia-php-sdk-v2/issues/101

typhonius:
I've just committed a change that should fix this if you want to kick off CI again.

grasmash:
Roger that.

.travis.yml
```diff
@@ -54,6 +55,7 @@ jobs:
 before_install:
   # Tests for the SSH key commands rely on ssh-agent running.
   - eval "$(ssh-agent -s)"
+  - composer self-update --preview
```
grasmash, line 58: This updates to Composer 2.

grasmash:
@danepowell you good with this?

composer.json
```diff
@@ -36,7 +36,7 @@
     "symfony/validator": "^5.1",
     "symfony/yaml": "^5.1@dev",
     "typhonius/acquia-logstream": "dev-master",
-    "typhonius/acquia-php-sdk-v2": "^2.0.14",
+    "typhonius/acquia-php-sdk-v2": "dev-master as 2.0.14",
```
danepowell, line 39: Looks like 2.0.15 is out now, I'd prefer using `^2.0.15` since aliases lead to weird dependency conflicts.

danepowell (approved):
Would prefer a stable dependency on the sdk but not a huge deal. Otherwise looks great.

PR #322, opened 2020-10-19
Fixes #277: padraic/phar-updater is abandoned.

### LINKED ISSUE #277: padraic/phar-updater is abandoned
On `composer install`:
> Package padraic/phar-updater is abandoned, you should avoid using it. No replacement was suggested.

I don't think this impacts end users, but obviously if this package is abandoned it could stability implications for us over time. Unfortunately I see no replacement. I doubt we want to maintain phar-updater, though that's one possible solution.

https://github.com/humbug/phar-updater/issues/59

grasmash:


danepowell:
I've confirmed via manual testing that this works as expected. However, we lose the `--allow-unstable` flag (probably not a big deal now that we only have stable releases), and we lose the update notification. We might be able to restore that.

anavarre:
The update notification seems pretty important to keep as it's what prevents us from having version fragmentation.

danepowell:
Alright I restored update checking. I just need to fix tests, but you can start manually testing.

Note the steps to test manually: https://github.com/acquia/cli/blob/845afb9799bc06d374836e55c9784a6cf18d4798/CONTRIBUTING.md

grasmash:
If the latest release is an alpha or beta, this will prompt the user to update. It doesn’t seem like desired functionality.

src/Command/CommandBase.php
```diff
@@ -1055,6 +1066,72 @@ protected function deleteSshKeyFromCloud(\stdClass $cloud_key): void {
     }
   }
 
+  /**
+   * @param \Symfony\Component\Console\Input\InputInterface $input
+   * @param \Symfony\Component\Console\Output\OutputInterface $output
+   */
+  protected function checkForNewVersion(InputInterface $input, OutputInterface $output): void {
+    try {
+      if (strpos($input->getArgument('command'), 'api:') === FALSE && $this->hasUpdate()) {
```
grasmash, line 1075: Why don’t we check for an update when running API commands?

src/Command/CommandBase.php
```diff
@@ -1055,6 +1066,72 @@ protected function deleteSshKeyFromCloud(\stdClass $cloud_key): void {
...
+        $output->writeln("A newer version of Acquia CLI is available. Run <comment>acli self-update</comment> to update.");
```
grasmash, line 1076: We should use the bold style rather than comments to show the example command. That’s the precedent that we have set elsewhere.

src/Command/CommandBase.php
danepowell: I have no idea, this is how you coded the original update command 😄 https://github.com/acquia/cli/pull/284/commits/2a38befb7f4c27e4ce57d139b1e9f2d6211fee59

danepowell:
> If the latest release is an alpha or beta, this will prompt the user to update. It doesn’t seem like desired functionality.

@grasmash indeed the consolidation/self-update library has no concept of stability or channels. Are you satisfied if we only support stable versions (i.e. x.y.z, with no alpha/beta/rc appended)?

anavarre:
I'm not @grasmash but I think we only care about stable releases at this point.

src/Command/CommandBase.php
grasmash: I did think it might have been me. Oh wait! I think it's so that we don't break the JSON output of api:* commands. Let's add a comment to that effect so that we don't forget.

grasmash:
I think keeping it stable is good. I think you can use _something like_ this to determine stability:
```
if (VersionParser::parseStability($version) === 'stable') {
```

danepowell:
Okay I changed it to only alert for stable releases, and tested the logic to ensure updates still appear only for newer releases (but I can't really test the stable/unstable logic).

I think we should follow up to add unstable channel support to consolidation/self-update, I'd imagine it's pretty simple (famous last words) and I'll try to open a PR.

danepowell:
Ugh, I just realized this fails if the most _recent_ release is unstable, but there's an _intermediate_ stable release between that and the version being run.

Why is this so hard?? It's almost like there's a reason package all of this logic as libraries (that then get abandoned... :smile: )

Since we don't plan on using unstable releases again in the near future, perhaps we should merge this as-is and fix this upstream in https://github.com/consolidation/self-update/issues/4

PR #326, opened 2020-10-19
DX-1752: Use single input/output objects for testing

danepowell:
@grasmash you opened this ticket, this might be halfway to a solution but I'm not quite sure where to take it from here.

I understand from an aesthetic perspective why having one input/output pair defined on the test base and another defined in each command is ugly, but functionally I don't see the problem.

The best solution I see is to set a default intput/output during setup (to populate all of the helper objects), and then reset them if necessary per test case. So for instance, the TelemetryCommandTest could call `setIo($new_input, $new_output)`.

But I guess I don't understand what you win here. It doesn't look like the Symfony Command tester even uses an actual input stream, it only takes a simple array.

Let me know what you want to do with this.

### DIFF
--- a/tests/phpunit/src/TestBase.php
+++ b/tests/phpunit/src/TestBase.php
@@ -177,14 +177,12 @@ abstract class TestBase extends TestCase {
     if (!$output) {
       $output = new BufferedOutput();
     }
+    $input = new ArrayInput([]);
 
     $this->application = new Application();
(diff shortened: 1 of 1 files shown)


grasmash:
I filed this ticket so that it would be possible to test the output of helper classes, like the TelemetryHelper. Previously we could not write tests like that because it would not receive the proper input arguments or output detectible text values.

PR #327, opened 2020-10-20
DX-1750: pull:files from multisite apps

danepowell:
Motivation
----------
Running `acli pull:files` does not let you choose a specific site to pull from. It just pulls from the default site.

Proposed changes
---------
Let the user choose which site to pull files from.

Alternatives considered
---------
None, we have to support multisite.

Testing steps
---------
Ensure that `acli pull:files` works when selecting a multisite, and (still) works for ACSF and the default site.


src/Command/CommandBase.php
```diff
@@ -1155,4 +1155,98 @@ protected function convertEnvironmentAliasToUuid(InputInterface $input): void {
...
+
+    return $sitegroup;
+  }
+
+  /**
+   * @param $cloud_environment
+   *
+   * @return bool
+   */
+  protected function isAcsfEnv($cloud_environment): bool {
```
danepowell, line 1175: These functions are not specific to pushing/pulling files, and could be useful more broadly, so I moved them here.

tests/phpunit/src/CommandTestBase.php
```diff
@@ -228,13 +228,45 @@ public function mockAcsfEnvironmentsRequest(
     return $response;
   }
 
+  /**
+   * @param object $applications_response
+   *
+   * @return object
+   * @throws \Psr\Cache\InvalidArgumentException
+   */
+  public function mockCloudEnvironmentsRequest(
```
danepowell, line 237: Apparently we were only mocking ACSF requests, leading to a lot of uncovered code paths especially in the push/pull commands (wherever `isAcsf()` was checked). This should make it easier to test those code paths (we'll still need to write new test cases for them).

src/Command/CommandBase.php
```diff
@@ -1155,4 +1155,98 @@ protected function convertEnvironmentAliasToUuid(InputInterface $input): void {
...
+
+  /**
+   * @param \AcquiaCloudApi\Response\EnvironmentResponse $cloud_environment
+   *
+   * @return array
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function getCloudSites($cloud_environment): array {
+    $sitegroup = self::getSiteGroupFromSshUrl($cloud_environment);
+    $command = ['cat', "/var/www/site-php/$sitegroup.{$cloud_environment->name}/config.json"];
```
grasmash, line 1212: I had no idea this file existed. How does it get populated?

src/Command/CommandBase.php
```diff
@@ -1155,4 +1155,98 @@ protected function convertEnvironmentAliasToUuid(InputInterface $input): void {
...
+
+  /**
+   * @param \AcquiaCloudApi\Response\EnvironmentResponse $cloud_environment
+   *
+   * @return mixed
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function promptChooseCloudSite($cloud_environment) {
+    $cloud_sites = $this->getCloudSites($cloud_environment);
+    $choices = array_keys($cloud_sites['databases']);
```
grasmash, line 1248: Are we assuming that a database == a site here?

src/Command/Pull/PullCommandBase.php
```diff
@@ -452,11 +452,15 @@ protected function rsyncFilesFromCloud($chosen_environment, $output_callback = N
...
+      $site = $this->promptChooseAcsfSite($chosen_environment);
       $source_dir = '/mnt/files/' . $sitegroup . '.' . $chosen_environment->name . '/sites/g/files/' . $site . '/files';
     }
     else {
-      $source_dir = '/home/' . $sitegroup . '/' . $chosen_environment->name . '/sites/default/files';
+      $site = $this->promptChooseCloudSite($chosen_environment);
+      if ($site == $sitegroup) {
+        $site = 'default';
+      }
+      $source_dir = '/mnt/files/' . $sitegroup . '.' . $chosen_environment->name . '/sites/' . $site . '/files';
```
grasmash, line 463: Will the database name always correspond this path?

src/Command/CommandBase.php
anavarre: `config.json` is not a valid path on Cloud IDE. The only path that exists ATM is `/var/www/site-php/SITEGROUP/DBNAME-settings.php`

src/Command/CommandBase.php
```diff
@@ -1155,4 +1155,98 @@ protected function convertEnvironmentAliasToUuid(InputInterface $input): void {
...
+
+  /**
+   * @param \AcquiaCloudApi\Response\EnvironmentResponse $cloud_environment
+   *
+   * @return array
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function getAcsfSites($cloud_environment): array {
+    $sitegroup = self::getSiteGroupFromSshUrl($cloud_environment);
+    $command = ['cat', "/var/www/site-php/$sitegroup.{$cloud_environment->name}/multisite-config.json"];
```
anavarre, line 1196: Not a valid path on Cloud IDE.

anavarre (changes requested):

src/Command/CommandBase.php
danepowell: `config.json` is part of the standard set of server-side includes on Cloud. It's not officially documented but AFAICT it's standard for every application, just like `multisite-config.json` is for ACSF.

We're calling this function on a `$cloud_environment` so it's fine that this file doesn't exist in IDEs.

src/Command/CommandBase.php
danepowell: Basically yes, a database is a site. This is enforced on the Cloud backend. The only exception to this is the "default" site, it's called by the sitegroup name in Cloud but "default" in Drupal.

We can either `ls` the site directories or read them from `config.json`. I chose to use `config.json` since that aligns with how we get metadata on ACSF apps.

danepowell: Oh... I think you are onto something. I just created a new DB in the Cloud UI and no corresponding directory was created on disk, so it looks like the directory structure mirrors what's in the codebase, not what's in Cloud config.

Ideally this is captured somewhere machine readable, otherwise we can just read the directories from disk as well.

src/Command/CommandBase.php
grasmash: @anavarre I don't think this is looking for that path on Cloud IDE. It's looking for that path on a Cloud Platform env.

src/Command/CommandBase.php
grasmash: I don't think it's captured anywhere. I'd suggest using the Symfony Finder component to list the dirs in `docroot/sites`. Something like BLT has:

```
  /**
   * Gets an array of sites for the Drupal application.
   *
   * I.e., sites under docroot/sites, not including acsf 'g' pseudo-sites.
   *
   * @return array
   *   An array of sites.
   */
  protected function getSiteDirs() {
    $sites_dir = $this->get('docroot') . '/sites';
    $sites = [];

    // If BLT's template has not yet been rsynced into the project root, it is
    // possible that docroot/sites does not exist.
    if (!file_exists($sites_dir)) {
      return $sites;
    }

    $finder = new Finder();
    $dirs = $finder
      ->in($sites_dir)
      ->directories()
      ->depth('< 1')
      ->exclude(['g', 'settings'])
      ->sortByName();
    foreach ($dirs->getIterator() as $dir) {
      $sites[] = $dir->getRelativePathname();
    }

    return $sites;
  }
```

src/Command/CommandBase.php
grasmash: Doesn't matter! :p

danepowell:
Alright I think this is good to go functionally, but I still need to get tests passing. Will finish it tomorrow.

danepowell:
I realized that if we're pulling files from multisite, we probably want to support pushing as well! Fixed that, and also cleaned up related tests. Tons of new code coverage here!

anavarre:
Tested the PR and it does look to be working as advertised :+1:

danepowell:
@grasmash can you review?

src/Command/CommandBase.php
```diff
@@ -1147,6 +1147,98 @@ protected function convertEnvironmentAliasToUuid(InputInterface $input): void {
...
+
+  /**
+   * @param \AcquiaCloudApi\Response\EnvironmentResponse $cloud_environment
+   *
+   * @return array
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function getCloudSites($cloud_environment): array {
+    $sitegroup = self::getSiteGroupFromSshUrl($cloud_environment);
+    $command = ['ls', "/mnt/files/$sitegroup.{$cloud_environment->name}/sites"];
```
grasmash, line 1204: Have you verified that this works on both CDEs and ACE sites?

danepowell: Yep I've tested on both CDE and ACE

PR #331, opened 2020-10-21
Add support for Lando db.

grasmash:


danepowell (approved):
Nice, looks simpler than expected. Do you want to take this opportunity to hide these commands in non-Lando / non-IDE envs?

anavarre:
If we don't hide those commands, people will ask about DDEV, Docksal, etc. support.

grasmash:
I've hidden them for all but IDE and Lando envs.

PR #332, opened 2020-10-22
Log Exception events immediately.

grasmash:


grasmash:
These failures reveal that this feature was never really working. We have to make sure that amplitude is correctly initialized very early to ensure that it’s available when an exception is thrown

nielsvm (approved):

nielsvm:
Although I approved, Travis reports utest failures, is that expected?

PR #333, opened 2020-10-22
Fix heisenbug in IDE create test

danepowell:
Motivation
----------
Kept getting this error running tests locally:
> 1) Acquia\Cli\Tests\Commands\Ide\IdeCreateCommandTest::testCreate
array_map(): Expected parameter 2 to be an array, null given

Seemed to be a bit of a heisenbug, it would come and go on different machines. I've never seen it in Travis CI, not sure why. Maybe there's an actual network call happening here? :grimacing: 

Proposed changes
---------
Mock environment requests for all applications in this test. This is necessary because the command loops through all applications, but we only mock the first one:
- https://github.com/acquia/cli/blob/1.0.0/src/Command/CommandBase.php#L488
- https://github.com/acquia/cli/blob/1.0.0/tests/phpunit/src/TestBase.php#L440

Alternatives considered
---------
I have no idea why this only happens in this one test, or why it happens intermittently. Quite possibly I'm missing something.

Testing steps
---------
Run `composer test` on a machine where you've previously seen this bug.

### DIFF
--- a/tests/phpunit/src/Commands/Ide/IdeCreateCommandTest.php
+++ b/tests/phpunit/src/Commands/Ide/IdeCreateCommandTest.php
@@ -23,8 +23,9 @@ class IdeCreateCommandTest extends CommandTestBase {
    */
   public function testCreate(): void {
 
-    $this->mockApplicationsRequest();
+    $applications_response = $this->mockApplicationsRequest();
     $this->mockApplicationRequest();
+    $this->mockAllEnvironmentsRequest($applications_response);
 
     // Request to create IDE.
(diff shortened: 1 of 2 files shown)


danepowell:
Oh great, now it passes every time locally but not on Travis :roll_eyes: 

Ya know... given that the environment request happens in the function that matches the local Git repo to a Cloud app, I bet this is happening because the PHPUnit tests are run in the acli repo locally. ACLI is trying to match _itself_ to a Cloud app when running locally.

danepowell:
Yeah I was way off initially, we just needed to respond "no" to the question to link repos.

PR #334, opened 2020-10-23
Fix #242: Support multisite db refresh

### LINKED ISSUE #242: To support different DBs in IDEs, Acquia CLI needs to create the corresponding <DB>-settings.inc file upon doing a refresh
Take an example. Sample IDE is provisioned with sitegroup "test". Automatically, the configuration overrides in the IDE will be stored under `/var/www/site-php/test/test-settings.inc`

But upon selecting a different database than the default one via Acquia CLI, then the expected path might become `/var/www/site-php/test/db2-settings.inc`

But the IDE in question won't be updated accordingly. This leaves the user in a dead end because the configuration overrides AND DB credentials are not preconfigured for this specific sitegroup's database.

What we should do is upon doing an `acli refresh` AND selecting a DB, rename `/var/www/site-php/<SITEGROUP>/<DB>-settings.inc` so `DB-settings.inc` is modified accordingly to the database name we're pulling.

danepowell:
Motivation
----------
Fixes #242

Proposed changes
---------
Upon a DB pull, copy the default db settings file to match the name of the pulled db.

Alternatives considered
---------
Tons, see #242 

Testing steps
---------
`acli pull:db` on a multisite app, ensure that a new file is created at `/var/www/site-php/[sitegroup]`

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -825,9 +825,8 @@ abstract class PullCommandBase extends CommandBase {
 
   /**
    * @param \AcquiaCloudApi\Response\EnvironmentResponse $source_environment
-   * @param \AcquiaCloudApi\Response\DatabaseResponse $database
    */
-  protected function createDbSettingsFile(EnvironmentResponse $source_environment, DatabaseResponse $database): void {
+  protected function createDbSettingsFile(EnvironmentResponse $source_environment, $database): void {
     $sitegroup = self::getSiteGroupFromSshUrl($source_environment);
     $default_path = "/var/www/site-php/$sitegroup/$sitegroup-settings.inc";
     $db_name_path = "/var/www/site-php/$sitegroup/{$database->name}-settings.inc";
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -8,7 +8,6 @@ use Acquia\Cli\Output\Checklist;
 use Acquia\DrupalEnvironmentDetector\AcquiaDrupalEnvironmentDetector;
 use AcquiaCloudApi\Connector\Client;
 use AcquiaCloudApi\Endpoints\Environments;
-use AcquiaCloudApi\Response\DatabaseResponse;
 use AcquiaCloudApi\Response\EnvironmentResponse;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputInterface;
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -89,11 +89,9 @@ abstract class PullCommandBase extends CommandBase {
     $this->checklist->addItem('Importing Drupal database copy from the Cloud Platform');
     $this->importRemoteDatabase($source_environment, $database, $this->getOutputCallback($output, $this->checklist));
     $sitegroup = self::getSiteGroupFromSshUrl($source_environment);
-    $db_name = $this->getNameFromDatabaseResponse($database);
     $default_path = "/var/www/site-php/$sitegroup/$sitegroup-settings.inc";
-    $db_name_path = "/var/www/site-php/$sitegroup/$db_name-settings.inc";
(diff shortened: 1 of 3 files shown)


danepowell:
@anavarre unfortunately this requires `sudo` because the files and directory are owned by root. We'll need to change that on the IDE side first, or else have ACLI ask for sudo perms. I think it's fine for IDEs to make these files world-writable since it's a development environment.

src/Command/Pull/PullCommandBase.php
```diff
@@ -88,6 +88,10 @@ protected function pullDatabase(InputInterface $input, OutputInterface $output):
     $database = $this->determineSourceDatabase($acquia_cloud_client, $source_environment);
     $this->checklist->addItem('Importing Drupal database copy from the Cloud Platform');
     $this->importRemoteDatabase($source_environment, $database, $this->getOutputCallback($output, $this->checklist));
+    $sitegroup = self::getSiteGroupFromSshUrl($source_environment);
+    $default_path = "/var/www/site-php/$sitegroup/$sitegroup-settings.inc";
+    $db_name_path = "/var/www/site-php/$sitegroup/{$database->name}-settings.inc";
```
grasmash, line 93: I think it would be better to use the symfony file system to do this.

anavarre:
> @anavarre unfortunately this requires sudo because the files and directory are owned by root. We'll need to change that on the IDE side first, or else have ACLI ask for sudo perms. I think it's fine for IDEs to make these files world-writable since it's a development environment.

Shouldn't be an issue because `acli` is owned by root in IDEs.

anavarre:
FTR I did try this PR today and the file wasn't created. I know it's a WIP, still, but wanted to give it a shot. Will try again when you say it's ready.

danepowell:
Hmm, I don't think it matters who owns the acli binary, it only matters (for the purposes of managing files) who _invokes_ the binary. That's going to be a non-privileged user. Pretty sure this won't work unless we change the settings file perms across all containers.

danepowell:
I opened a corresponding Cloud IDE PR to fix the perms on the sitegroup settings directory.

anavarre:
Yes, I think you're right @danepowell. I tried the PR against an existing IDE and ran into a permissions issue. When I have `/var/www/site-php/<sitegroup>` owned by the IDE user, then I can successfully see the file is being created.

src/Command/Pull/PullCommandBase.php
```diff
@@ -88,6 +89,7 @@ protected function pullDatabase(InputInterface $input, OutputInterface $output):
     $database = $this->determineSourceDatabase($acquia_cloud_client, $source_environment);
     $this->checklist->addItem('Importing Drupal database copy from the Cloud Platform');
     $this->importRemoteDatabase($source_environment, $database, $this->getOutputCallback($output, $this->checklist));
+    $this->createDbSettingsFile($source_environment, $database);
```
grasmash, line 92: You should verify that this is an IDE since it is possible to run this command elsewhere.

grasmash (changes requested):
Please resolve comment(s).

src/Command/Pull/PullCommandBase.php
danepowell: Ah right, forgot about Lando...

tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
```diff
@@ -52,6 +45,19 @@ public function testPullDatabases(): void {
...
 
+  /**
+   * Test that settings files are created for multisite DBs in IDEs.
+   *
+   * @throws \Exception
+   */
+  public function testPullDatabaseSettingsFiles(): void {
+    $this->setupPullDatabase(TRUE, TRUE, TRUE, TRUE, TRUE, TRUE);
+    $inputs = $this->getInputs();
+    putenv('AH_SITE_ENVIRONMENT=IDE');
```
danepowell, line 56: I have to set this variable manually instead of using the IDE test base, because the dedicated test base also sets an app ID and other variables that seem to conflict with the rest of this test.

anavarre:
I tested the latest changes in this PR and it still works fine. I wonder whether we should add a new step in the `pull:` operation to indicate we're creating a new settings file, and possibly indicate the location, since it's user-editable?

PR #348, opened 2020-11-06
Replace --cloud-env-uuid option with environmentId arg.

grasmash:
Using the environmentId argument name allows users to use either an alias or a uuid.


danepowell (approved):
Cool!

PR #350, opened 2020-11-09
Throw exception after 30 min IDE DNS timeout.

grasmash:


nielsvm:
https://backlog.acquia.com/browse/DX-2732

src/Command/Ide/IdeCreateCommand.php
```diff
@@ -96,7 +101,7 @@ protected function waitForDnsPropagation($ide_url): void {
         $this->logger->debug($e->getMessage());
       }
     });
-    LoopHelper::addTimeoutToLoop($loop, 15, $spinner, $this->output);
+    LoopHelper::addTimeoutToLoop($loop, 20, $spinner, $this->output);
```
nielsvm, line 104: Why 20 minutes? I'm not aware of the average but wouldn't 15 suffice?

src/Helpers/LoopHelper.php
```diff
@@ -47,8 +48,8 @@ public static function addTimeoutToLoop(
   ): void {
     $loop->addTimer($minutes * 60, function () use ($loop, $minutes, $spinner, $output) {
       self::finishSpinner($spinner);
-      $output->writeln("Timed out after $minutes minutes!");
       $loop->stop();
+      throw new AcquiaCliException("Timed out after $minutes minutes!");
```
nielsvm, line 52: This would skip printing the output URL right?

nielsvm (approved):

src/Command/Ide/IdeCreateCommand.php
anavarre: 15mn often fails for people not using Google or Cloudflare DNS servers. 20 is just buying us extra time, but won't fix the issue with DNS propagation times in Australia, India and parts of EMEA.

nielsvm: That's good context to know, thanks.

src/Helpers/LoopHelper.php
nielsvm: Confirmed my own question:
https://github.com/acquia/cli/pull/350/files/ff4887ad59953beac33f64417417273226f214f0#diff-dc099926d287b14cf26be796bb6b26f83b5c8541ff2b9f78dbb358871f1ecdf9R107

anavarre: I'm on the fence about doing this. Sometimes the timeout happens but URLs are actually accessible a few seconds/minutes after. I'd rather have a better message that explains it's not us, it's the DNS system, and the 2 URLs will shortly be available.

nielsvm: Its a failure nonetheless, and I think we should also consider removing the just created IDE if it expired DNS.

danepowell:
Maybe also a message to the effect of "Acquia has been alerted to this issue, please follow status.acquia.com for updates."

anavarre:
> Maybe also a message to the effect of "Acquia has been alerted to this issue, please follow status.acquia.com for updates."

Which wouldn't be true. You can have a DNS propagation issue on your end, or with the DNS system entirely. That wouldn't be an outage on our end nor would indicate status.acquia.com would necessarily be updated.

nielsvm:
@danepowell yeah that's what I suggested in the filed ticket:
https://backlog.acquia.com/browse/DX-2732

nielsvm:
@anavarre yeah so maybe we shouldn't be leveraging DNS checking from a users machine but from our own heart-beating logic?

anavarre:
@nielsvm pre-provisioning should no longer have us run into the issue unless exceptional circumstances.

nielsvm:
@grasmash how do you feel about relaying to `IdeDeleteCommand->execute()` as part of the exception handler for `CreateCommand`? I've got a strong feeling we shouldn't ever leave behind half-baked IDEs even if pre-provisioning will mask much of the issue here.

grasmash:
I don’t feel like that’s the responsibility of the CLI tool. I agree that we should not leave around half baked IDEs, but that’s something that should be taken care of on the service side rather than the client side.

src/Helpers/LoopHelper.php
grasmash: I'v changed this to catch the exception and print an error followed by the IDE links.

danepowell (approved):

PR #359, opened 2020-11-17
Fork consolidation/self-update until PR is merged

danepowell:
I think this upstream PR will be merged soon, but until it is, @anavarre and I think it's best to fork that repo so we can get a new release out with a full backwards-compatible self-update command.

https://github.com/consolidation/self-update/pull/5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,18 @@
     }
   ],
   "minimum-stability": "dev",
+  "repositories": [
+    {
+      "type": "vcs",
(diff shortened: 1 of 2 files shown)


danepowell:
Send commit hash to Aurelien for review after merging.

anavarre (approved):
I haven't tested this nor do I know how I could test it. If you have steps for me to follow, I'm happy to do just that.

Otherwise, I understand what was done here and think it's the right way forward until the PR is merged upstream.

PR #360, opened 2020-11-19
Disable update notifications in development

danepowell:
When developing on ACLI you get an update notification on every command because of the placeholder version string.


anavarre:
Without this PR applied:

```
$ acli telemetry
A newer version of Acquia CLI is available. Run acli self-update to update.
Telemetry has been enabled.
Run this command again to disable telemetry
```

With the PR applied:

```
$ acli telemetry
Telemetry has been disabled.
Run this command again to enable telemetry
```

PR #363, opened 2020-11-20
DX-2641: Allow passing SSH commands

danepowell:
Motivation
----------
@eporama requested the ability to run custom commands via `remote:ssh`

Proposed changes
---------
Add the ability to pass custom commands. If no command is provided, default to the old behavior (open a shell).

Testing steps
---------
3. Run `acli remote:ssh myapp.dev` to ensure the original behavior still works.
4. Run `acli remote:ssh myapp.dev -- ls -al` or any other command to see that it gets executed via SSH.

### DIFF
--- a/src/Command/Remote/SshCommand.php
+++ b/src/Command/Remote/SshCommand.php
@@ -20,11 +20,12 @@ class SshCommand extends SshBaseCommand {
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription('Open a new SSH session to a Cloud Platform environment')
+    $this->setDescription('Use SSH to open a shell or run a command in a Cloud Platform environment')
       ->setAliases(['ssh'])
       ->addArgument('alias', InputArgument::REQUIRED, 'Alias for application & environment in the format `app-name.env`')
-      ->addArgument('ssh_command', InputArgument::OPTIONAL, 'Command to run via SSH (opens a shell by default)')
-      ->addUsage("<app>.<env>");
+      ->addArgument('ssh_command', InputArgument::OPTIONAL, 'Command to run via SSH (if not provided, opens a shell in the site directory)')
+      ->addUsage("myapp.dev # open a shell in the myapp.dev environment")
+      ->addUsage("myapp.dev ls # list files in the myapp.dev environment and return");
   }
 
   /**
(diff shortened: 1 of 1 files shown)


grasmash:
I think we should update the command description to explain that this will either open a shell or alternatively execute a command remotely. We should also add multiple usage examples

anavarre:
Manual test

```
$ acli remote:ssh myapp.dev "ls -l"
total 0
lrwxrwxrwx 1 root       root       23 Dec 10  2015 dev -> /mnt/gfs/myapp.dev
drwxr-xr-x 6 myapp myapp 92 Jun  8  2016 drush-backups
lrwxrwxrwx 1 root       root       24 Dec 10  2015 test -> /mnt/gfs/myapp.test
Connection to myappdev.ssh.prod.acquia-sites.com closed.
$ acli remote:ssh myapp.dev "stat -c "%U" drush-backups"
anavarred8
Connection to myappdev.ssh.prod.acquia-sites.com closed.
```

anavarre (approved):

src/Command/Remote/SshCommand.php
```diff
@@ -20,11 +20,12 @@ class SshCommand extends SshBaseCommand {
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription('Open a new SSH session to a Cloud Platform environment')
+    $this->setDescription('Use SSH to open a shell or run a command in a Cloud Platform environment')
       ->setAliases(['ssh'])
       ->addArgument('alias', InputArgument::REQUIRED, 'Alias for application & environment in the format `app-name.env`')
-      ->addArgument('ssh_command', InputArgument::OPTIONAL, 'Command to run via SSH (opens a shell by default)')
-      ->addUsage("<app>.<env>");
+      ->addArgument('ssh_command', InputArgument::OPTIONAL, 'Command to run via SSH (if not provided, opens a shell in the site directory)')
```
grasmash, line 26: I think we may actually want to may this an array argument, otherwise users will have to encapsulate the command in quotes. See https://github.com/acquia/cli/blob/master/src/Command/Remote/DrushCommand.php#L27

In that case, the example could be something like `acli ssh -- ls -lash`

PR #365, opened 2020-11-25
DX-2666: Test all exception listeners

danepowell:
Motivation
----------
Only one exception listener is being tested right now, and it's not scalable to test other exception listeners like that because the existing test bootstraps the whole application.

Proposed changes
---------
Unit test _all_ exception listeners using a unit test (no bootstrap). Also move misc test cases into a separate folder from the abstract bases, just for organization's sake.

Alternatives considered
---------
Build the whole application to test listeners end-to-end, but this is going to be really ugly and slow, and doesn't add much in the way of coverage.

Testing steps
---------

1. Just verify that code coverage goes up.


anavarre:
How can I test this?

danepowell:
There's nothing to manually test since the only changes are to test cases, and the only effect is to increase test coverage. So just look at the automated checks on this PR and verify that tests are passing and coverage increased.

anavarre (approved):

PR #374, opened 2020-12-02
Fixes #373: pull:files does not work for Polaris environments.

### LINKED ISSUE #373: pull:files does not work for Polaris environments
Currently Acquia CLI looks in `/mnt/files` to list Cloud sites via promptChooseCloudSite(). This directory does not exist on Polaris. Instead, we must check `/mnt/gfs`

grasmash:


grasmash:
Depends on https://github.com/typhonius/acquia-php-sdk-v2/pull/118

PR #379, opened 2020-12-08
Fix #367: Support PHP 8.0

### LINKED ISSUE #367: Support PHP 8
ACLI would likely run fine on PHP 8 if we just relax the version constraint in composer.json, but we should verify this and then change the constraint.

Currently, trying to use PHP 8 produces:
```
Box Requirements Checker
========================
> Using PHP 8.0.0
> PHP is using the following php.ini file:
  /usr/local/etc/php/8.0/php.ini
> Checking Box requirements:
  E....
 [ERROR] Your system is not ready to run the application.
Fix the following mandatory requirements:
=========================================
 * The application requires the version "^7.3" or greater.
```

danepowell:
Motivation
----------
Fixes #367

Proposed changes
---------
- Add PHP 8 to composer.json
- Add PHP 8 to Travis tests

### DIFF
--- a/config/services.yml
+++ b/config/services.yml
@@ -54,7 +54,11 @@ services:
 
   # AcquiaCloudApi services.
   AcquiaCloudApi\Connector\Client: ~
-  AcquiaCloudApi\Connector\ConnectorInterface: ~
+  AcquiaCloudApi\Connector\Connector:
+    arguments:
+      $config: 'something'
+  AcquiaCloudApi\Connector\ConnectorInterface:
+    alias: AcquiaCloudApi\Connector\Connector
   AcquiaLogstream\LogstreamManager: ~
 
   # Symfony services.
--- a/config/services.yml
+++ b/config/services.yml
@@ -56,7 +56,7 @@ services:
   AcquiaCloudApi\Connector\Client: ~
   AcquiaCloudApi\Connector\Connector:
     arguments:
-      $config: 'something'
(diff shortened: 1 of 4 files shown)


config/services.yml
```diff
@@ -54,7 +54,11 @@ services:
 
   # AcquiaCloudApi services.
   AcquiaCloudApi\Connector\Client: ~
-  AcquiaCloudApi\Connector\ConnectorInterface: ~
+  AcquiaCloudApi\Connector\Connector:
+    arguments:
+      $config: ['@=service("datastore.cloud").get("key")', '@=service("datastore.cloud").get("secret")']
```
danepowell, line 59: I'm pretty sure that splitting this into an interface (ConnectorInterface) and subclass (Connector) defined as an alias is the right move, since it follows the pattern further down for Symfony Input and Output interfaces.

Manually defining the argument is necessary to fix [this test failure](https://travis-ci.com/github/acquia/cli/jobs/456247653#L3099). However: I'm pretty sure this is not working like we expect. Tests pass and everything works fine, but in my testing I could define _any_ value as this argument and it would still work.

Possibly this is because of how we use a factory method to produce a Client service, and that needs to be refactored too? I hate Symfony wiring 😬

danepowell: I think we should merge this as-is and we'll clean up the wiring in a follow-up issue: https://github.com/acquia/cli/pull/380

anavarre (approved):

PR #380, opened 2020-12-08
DX-3045: Remove ClientService Helper

danepowell:
Motivation
----------
Depends on #379. Still needs a lot of work (haven't actually remove the ClientService helper yet)

In #379 I discovered that our autowiring was broken in PHP 8. And in fact, we weren't even using the wiring related the AcquiaCloudApi client/connector, because of the ClientService factory that bypassed Symfony autowiring.

If we fix the autowiring, I think we can get rid of the ClientService helper completely.

Proposed changes
---------
Kill the ClientService helper and use autowiring to provide a Cloud API client to commands.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ version: ~> 1.0
 php:
   - 7.3
   - 7.4
+  - 8.0.0
 
 cache:
   directories:
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
     }
   ],
   "require": {
-    "php": "^7.3",
+    "php": "^7.3 | ^8.0",
     "ext-json": "*",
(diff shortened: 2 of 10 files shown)


danepowell:
After reading the acquia-php-sdk-v2 docs, I realize that a factory pattern for the client (as we've ostensibly done with the ClientService) is the right idea, but in fact we've screwed up in two ways:
- It's not a true factory since we are storing and re-using the client
- The Connector autowiring isn't working, making the service more complex than it needs to be.

We can either fix the ClientService or get rid of it entirely, either way a slightly different approach is needed. Storing and re-using the client as we've done here is wrong.

Follow-up: https://github.com/acquia/cli/pull/381

PR #381, opened 2020-12-10
DX-3045: Don't reuse AcquiaCloudApi client

danepowell:
Motivation
----------
We re-use the Acquia Cloud API client across multiple requests, which probably isn't good because we directly modify the client query and options in commands. This means that queries/options from one Cloud API call could bleed into the next, even across different commands. We have seen odd behavior from the Cloud API client that could be attributable to this, though I don't know for sure.

Proposed changes
---------
- Use a Client factory pattern as recommended in the acquia-php-sdk-v2 docs instead of re-using a single client
- Fix autowiring so that we can greatly simplify the ClientService class

Alternatives considered
---------
We could probably get rid of the ClientService completely, but it's still useful to facilitate testing. If we got rid of it, then we'd use dependency injection to store $connector on the command base, and in every `Command::execute()` we'd call `Client::factory($this->connector);`. This would break testing because it prevents us from prophesizing the client calls (see how `TestBase::clientProphecy` is used for mocking)

Testing steps
---------
3. Just make sure that commands that invoke the API (`ide:list`, `api:accounts:find`, etc) work as before

### DIFF
--- a/src/Helpers/ClientService.php
+++ b/src/Helpers/ClientService.php
@@ -4,32 +4,19 @@ namespace Acquia\Cli\Helpers;
 
 use AcquiaCloudApi\Connector\Client;
 use AcquiaCloudApi\Connector\Connector;
-use Webmozart\KeyValueStore\JsonFileStore;
 
 class ClientService {
 
-  private $acquiaCloudClient;
+  private $connector;
 
-  private $cloud_api_conf;
-
-  public function __construct(JsonFileStore $datastoreCloud) {
-    $this->cloud_api_conf = $datastoreCloud;
+  public function __construct(Connector $connector) {
+    $this->connector = $connector;
   }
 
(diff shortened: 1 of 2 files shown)


src/Helpers/ClientService.php
```diff
@@ -4,32 +4,19 @@
...
 class ClientService {
 
-  private $acquiaCloudClient;
+  private $connector;
 
-  private $cloud_api_conf;
-
-  public function __construct(JsonFileStore $datastoreCloud) {
-    $this->cloud_api_conf = $datastoreCloud;
+  public function __construct(Connector $connector) {
```
danepowell, line 12: Connector is passed automatically via dependency injection, thanks to the config/services.yml fix above

grasmash (approved):

PR #384, opened 2020-12-15
DX-2994: Fix #366: Log exceptions

### LINKED ISSUE #366: Exceptions are not properly logged
As proven by the failing test in https://github.com/acquia/cli/pull/332, our current implementation exception logging does not work.

It should be fixed, but it is not required that Amplitude be the solution. We can log errors in another service if that would be better.

danepowell:
Motivation
----------
Fixes #366

Proposed changes
---------
Use Amplitude consistently as a [singleton service](https://github.com/zumba/amplitude-php#zumba-amplitude-php) so that exceptions are reported correctly.

Alternatives considered
---------
Use Amplitude as an object or Symfony service, and pass it everywhere it's needed (including the exception). I don't know how to do this in a clean way.

Testing steps
---------

3. Remove your API credentials
4. Run a command requiring authentication (`ide:create`) and verify that the exception thrown shows up in Amplitude

screenshot-analytics amplitude com-2020 12 15-13_13_34

### DIFF
--- a/src/Command/Api/ApiCommandHelper.php
+++ b/src/Command/Api/ApiCommandHelper.php
@@ -15,7 +15,6 @@ use Symfony\Component\Console\Input\InputDefinition;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Yaml\Yaml;
 use Webmozart\KeyValueStore\JsonFileStore;
-use Zumba\Amplitude\Amplitude;
 
 /**
  *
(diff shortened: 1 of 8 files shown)


anavarre (approved):
Dane and I have confirmed my exceptions are logged correctly.

grasmash:
Seems like there's no test coverage?

danepowell:
I don't think we can test Amplitude calls because we can't mock the static Amplitude class / singleton service.

PR #396, opened 2021-01-16
DX-1682: Download db backups via API rather than SSH

danepowell:
This isn't quite complete (needs tests, code style cleanup) but it should be functional. I wanted to get it up before the long US weekend for initial feedback.

Instead of downloading db backups via MySQL dumps and SSH, this creates an on-demand backup via the Cloud API and then downloads the backup.

This is awesome because it's safer and allows more granular permissioning (no more SSH access to prod! 😬 ), but has two potential gotchas:
- It's a bit slower, because we have to wait for the notification api to tell us that the backup is ready.
- It's not deterministic, in that we have no way to ensure we are downloading the backup _we_ just created, since the notification API does not provide a backup ID. Instead, we just download the most recent backup on the assumption that it's ours.

### DIFF
--- a/src/Command/Pull/PullDatabaseCommand.php
+++ b/src/Command/Pull/PullDatabaseCommand.php
@@ -19,7 +19,8 @@ class PullDatabaseCommand extends PullCommandBase {
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription('Copy database from a Cloud Platform environment')
+    $this->setDescription('Import latest database backup from a Cloud Platform environment')
+      ->setHelp('This uses the latest available database backup, which may be up to 24 hours old. You can generate an on-demand backup using api:environments:database-backup-create.')
       ->setAliases(['pull:db'])
       ->addArgument('environmentId', InputArgument::OPTIONAL, 'The UUID of the associated Cloud Platform source environment')
       ->addOption('no-scripts', NULL, InputOption::VALUE_NONE,
--- a/tests/fixtures/backups_response.json
+++ b/tests/fixtures/backups_response.json
@@ -1,0 +1,12 @@
+[
+  {
+    "id": 1,
+    "database": "profserv2",
+    "type": "user",
+    "started_at": "something",
+    "completed_at": "something",
+    "flags": "",
+    "environment": "",
+    "_links": ""
+  }
+]
--- a/tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
+++ b/tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
@@ -8,6 +8,7 @@ use Acquia\Cli\Helpers\SshHelper;
 use AcquiaCloudApi\Response\EnvironmentResponse;
 use Prophecy\Argument;
 use Prophecy\Prophecy\ObjectProphecy;
+use Psr\Http\Message\StreamInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Filesystem\Filesystem;
 
@@ -51,7 +52,7 @@ class PullDatabaseCommandTest extends PullCommandTestBase {
    * @throws \Exception
    */
   public function testPullDatabaseSettingsFiles(): void {
-    $this->setupPullDatabase(TRUE, TRUE, TRUE, TRUE, TRUE, TRUE);
+    $this->setupPullDatabase(TRUE, TRUE, TRUE, TRUE, TRUE);
     $inputs = $this->getInputs();
     // @todo Use the IdeRequiredTestBase instead of setting AH_SITE_ENVIRONMENT.
     // IdeRequiredTestBase sets other env vars (such as application ID) that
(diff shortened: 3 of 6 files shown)


anavarre:
Manual tests:

```
$ time acli pull:database
Using Cloud Application testme

 Choose a Cloud Platform environment:
  [0] Dev, dev (vcs: pipelines-build-master)
  [1] Stage, test (vcs: tags/2020-10-19)
 > 0

 Choose a database [test (default)]:
  [0] test (default)
  [1] test2
 > 0

    ✔ Importing Drupal database copy from the Cloud Platform

real    0m17.172s
user    0m1.144s
sys     0m0.314s
$ time ./bin/acli pull:database
Using Cloud Application testme

 Choose a Cloud Platform environment:
  [0] Dev, dev (vcs: pipelines-build-master)
  [1] Stage, test (vcs: tags/2020-10-19)
 > 0

 Choose a database [test (default)]:
  [0] test (default)
  [1] test2
 > 0

    ✔ Waiting for database backup to complete...
    ✔ Importing Drupal database copy from the Cloud Platform
    ✔ Waiting for database backup to complete...

real    0m25.210s
user    0m1.299s
sys     0m0.314s
```

It's slower (I guess this is exacerbated with larged DB), but not only is this safer and will allow us to enable pulling from prod again since it'll back up the DB from the replica. Could we do this in this PR?

Also, I noticed a few display artifacts (not all visible above). Is there a way we can improve this?

grasmash:
Something to consider about this approach. It may create a large number of back ups for an application. Is there any chance that we will hit a maximum limit for back ups or a storage space?

grasmash:
I’m also not crazy about the nondeterministic aspect of this.

anavarre:
> I’m also not crazy about the nondeterministic aspect of this.

Me neither. Perhaps we need to enforce a change upstream in Cloud Platform API so this is supported.

> Something to consider about this approach. It may create a large number of back ups for an application. Is there any chance that we will hit a maximum limit for back ups or a storage space?

Nothing is ever infinite so yes, I guess. What about asking the user whether we should create an on-demand backup or pull the latest daily backup for the source environment in question? This way we might get a DB backup for free, because it'd already be there?

danepowell:
I like the idea of pulling the latest daily (not sure if we can actually tell what this is via the API, I'll check) so we can at least get this merged, and then we can follow up with additional on-demand functionality in coordination with the Cloud API team re: adding backup IDs to notifications.

danepowell:
I changed this to pull the latest available backup (daily at the latest), and created a follow-up ticket to integrate on-demand functionality if we so choose.

anavarre:
Manual test looks good. I am seeing the following:

```
[debug] Using database backup generated at 2021-01-21T07:12:03+00:00
```

Full log file:

```
[debug] Using database backup generated at 2021-01-21T07:12:03+00:00
[debug] Downloading database backup to /tmp/backup-2021-01-21T07:12:03+00:00-redacted.sql.gz
[debug] Dropping database drupal
mysql: [Warning] Using a password on the command line interface can be insecure.

[notice] Command: 'mysql' '--host' 'localhost' '--user' 'drupal' '--password=drupal' '-e' 'DROP DATABASE IF EXISTS drupal' [Exit: 0]
[debug] Creating new empty database drupal
mysql: [Warning] Using a password on the command line interface can be insecure.

[notice] Command: 'mysql' '--host' 'localhost' '--user' 'drupal' '--password=drupal' '-e' 'create database drupal' [Exit: 0]
[debug] Importing /tmp/backup-2021-01-21T07:12:03+00:00-redacted.sql.gz to MySQL on local machine
1.68MiB [ 320KiB/s]
    ✔ Importing Drupal database copy from the Cloud Platform
```

src/Command/Pull/PullDatabaseCommand.php
```diff
@@ -19,7 +19,8 @@ class PullDatabaseCommand extends PullCommandBase {
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription('Copy database from a Cloud Platform environment')
+    $this->setDescription('Import latest database backup from a Cloud Platform environment')
+      ->setHelp('This uses the latest available database backup, which may be up to 24 hours old. You can generate an on-demand backup using api:environments:database-backup-create.')
```
anavarre, line 23: As a user, when am I supposed to read this message?

tests/fixtures/backups_response.json
```diff
@@ -0,0 +1,12 @@
...
+    "id": 1,
+    "database": "profserv2",
+    "type": "user",
+    "started_at": "something",
+    "completed_at": "something",
+    "flags": "",
+    "environment": "",
+    "_links": ""
+  }
+]
```
anavarre, line 12: Do you actually need a new line at EOF?

tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
```diff
@@ -116,12 +106,14 @@ public function testPullDatabaseWithMySqlImportError(): void {
...
   }
 
-  protected function setupPullDatabase($mysql_dump_successful, $mysql_dl_successful, $mysql_drop_successful, $mysql_create_successful, $mysql_import_successful, $mock_ide_fs = FALSE): void {
+  protected function setupPullDatabase($mysql_dl_successful, $mysql_drop_successful, $mysql_create_successful, $mysql_import_successful, $mock_ide_fs = FALSE): void {
     $applications_response = $this->mockApplicationsRequest();
     $this->mockApplicationRequest();
     $environments_response = $this->mockAcsfEnvironmentsRequest($applications_response);
     $this->createMockGitConfigFile();
     $this->mockDatabasesResponse($environments_response);
+    $this->mockDatabaseBackupsResponse($environments_response, 'profserv2');
```
anavarre, line 115: Generally, shouldn't `profserv2` be added to a variable?

anavarre (approved):

src/Command/Pull/PullDatabaseCommand.php
danepowell: You'll see the help if you run `acli help pull:db`. I debated also making it part of the command output (when pulling the db) but that seemed too verbose.

tests/fixtures/backups_response.json
danepowell: AFAIK, best practice is to always end files with a new line. This is [explicitly true for Drupal](https://www.drupal.org/docs/develop/standards/coding-standards#indenting), not sure about other ecosystems.

tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
danepowell: Indeed it should, but I couldn't figure out where it was coming from in our mock responses. I've opened a follow-up ticket to clean up those mock responses to avoid issues like this.

PR #411, opened 2021-02-09
Mock application test network requests

danepowell:
**Motivation**
Application tests were making actual Cloud API requests, which made them somewhat flaky and impossible to mock or test for anything other than an unauthorized response. This also necessitated maintaining CA certs in Travis, which has been its own kind of trouble (https://github.com/acquia/cli/pull/408)

**Proposed changes**
Mock the request so that tests are faster and deterministic, and remove Travis CA certs so we don't have to maintain them. Use this pattern to extend test coverage with additional application tests, which is more robust than testing the ExceptionListener in isolation like we used to.

Mocking test services necessitated making all of our Symfony services public, which seems to be a known issue and best practice for CLI apps like ours that don't inherit from the Framework Bundle (and thus don't have access to the dedicated `test.service_container`).

**Testing steps**

3. Run `netstat` to see open connections.
4. `./vendor/bin/phpunit tests/phpunit/src --filter=ExceptionApplicationTest`
5. Run `netstat` to verify that there are no new connections.

**Merge requirements**


tests/phpunit/src/Application/ExceptionApplicationTest.php
```diff
@@ -18,26 +18,49 @@
  */
 class ExceptionApplicationTest extends TestBase {
 
+  protected $kernel;
```
nielsvm, line 21: Missing type docblock?

tests/phpunit/src/Application/ExceptionApplicationTest.php
```diff
@@ -18,26 +18,49 @@
...
+
   public function setUp($output = NULL):void {
     parent::setUp($output);
     // If kernel is cached from a previous run, it doesn't get detected in code
     // coverage reports.
     $this->fs->remove('var/cache');
+    $this->kernel = new Kernel('dev', 0);
```
nielsvm, line 28: Why moving the kernel initialization into `setUp` while it used only for only one test before? Anyway I'm not super opinionated, was just wondering...

tests/phpunit/src/Application/ExceptionApplicationTest.php
```diff
@@ -18,26 +18,49 @@
...
+    $this->kernel->boot();
+    $this->kernel->getContainer()->set('datastore.cloud', $this->datastoreCloud);
+    $this->kernel->getContainer()->set('Acquia\Cli\Helpers\ClientService', $this->clientServiceProphecy->reveal());
```
nielsvm, line 31:
```
$this->clientServiceProphecy->reveal()
```
The effect of this is basically the same as `getMock()` in the past right, in that it sets a shim-level functional mock of `Acquia\Cli\Helpers\ClientService`?

nielsvm (approved):
While working on Cloud IDEs this morning, I was first trying to mock out the network requests before basically giving up given how closely coupled the client is in the current code. This PR makes a lot of sense in that regard as it could serve more than just mocking inside the tests.

tests/phpunit/src/Application/ExceptionApplicationTest.php
danepowell: I'm not sure I understand. The Kernel needs to boot for every test case (since it holds the application container), so `setUp` seems the natural place to initialize it.

tests/phpunit/src/Application/ExceptionApplicationTest.php
danepowell: I think that's right. `clientServiceProphecy` is a prophet that lets you mock classes (in this case, `ClientService`), but it's not the mock/double class itself. You have to call `reveal()` to get the mock class.

PR #416, opened 2021-02-18
DX-3448 | Adding violuke/rsa-ssh-key-fingerprint and code to show fin…

eporama:
…gerprint hashes

**Motivation**

`acli ssh-key:list` shows a compact view of the keys returned by Cloud API and local, but does not provide the UUID which is necessary to be able to call `ssh-key:delete` with the `-uuid, --cloud-key-uuid=CLOUD-KEY-UUID` options.

It also does not provide a way to troubleshoot ssh connections by showing fingerprints of the keys.

Since CloudAPI fingerprint on the ssh key is MD5 but most logs and ssh output will note the fingerprint of the key as sha256 (see CXAPI-6802 for more info on getting sha256 added to the output), it would be handy to output these values as well.

**Proposed changes**
This simply changes the output of a listing command, no impact to users

**Merge requirements**

### DIFF
--- a/src/Command/Ssh/SshKeyListCommand.php
+++ b/src/Command/Ssh/SshKeyListCommand.php
@@ -34,7 +34,7 @@ class SshKeyListCommand extends SshKeyCommandBase {
     $cloud_keys = $acquia_cloud_client->request('get', '/account/ssh-keys');
     $local_keys = $this->findLocalSshKeys();
 
-    $this->io->title('Cloud keys with matching local keys');
+    $this->io->title('Cloud Platform keys with matching local keys');
     foreach ($local_keys as $local_index => $local_file) {
       foreach ($cloud_keys as $index => $cloud_key) {
         if (trim($local_file->getContents()) === trim($cloud_key->public_key)) {
@@ -44,14 +44,14 @@ class SshKeyListCommand extends SshKeyCommandBase {
             ['UUID' => $cloud_key->uuid],
             ['Local filename' => $local_file->getFilename()],
             ['sha256 hash' => $sha256_hash],
-            ['md5' => $cloud_key->fingerprint]
+            ['md5 hash' => $cloud_key->fingerprint]
           );
           unset($cloud_keys[$index], $local_keys[$local_index]);
           break;
         }
       }
     }
-    $this->io->title('Cloud keys with no matching local keys');
+    $this->io->title('Cloud Platform keys with no matching local keys');
     foreach ($cloud_keys as $index => $cloud_key) {
       $sha256_hash = FingerprintGenerator::getFingerprint($cloud_key->public_key, 'sha256');
       $this->io->definitionList(
@@ -59,11 +59,11 @@ class SshKeyListCommand extends SshKeyCommandBase {
         ['UUID' => $cloud_key->uuid],
         ['Local filename' => 'none'],
         ['sha256 hash' => $sha256_hash],
-        ['md5' => $cloud_key->fingerprint]
+        ['md5 hash' => $cloud_key->fingerprint]
       );
     }
 
-    $this->io->title('Local keys with no matching cloud keys');
+    $this->io->title('Local keys with no matching Cloud Platform keys');
     foreach ($local_keys as $index => $local_file) {
       $sha256_hash = FingerprintGenerator::getFingerprint($local_file->getContents(), 'sha256');
       $md5_hash = FingerprintGenerator::getFingerprint($local_file->getContents(), 'md5');
@@ -72,7 +72,7 @@ class SshKeyListCommand extends SshKeyCommandBase {
         ['UUID' => 'none'],
         ['Local filename' => $local_file->getFilename()],
         ['sha256 hash' => $sha256_hash],
-        ['md5' => $md5_hash]
+        ['md5 hash' => $md5_hash]
       );
     }
 
--- a/src/Command/Ssh/SshKeyListCommand.php
+++ b/src/Command/Ssh/SshKeyListCommand.php
@@ -41,19 +41,19 @@ class SshKeyListCommand extends SshKeyCommandBase {
       'Hashes — sha256 and md5',
     ]);
     // First list all keys for which there are both Cloud and local keys.
+    $last_key = array_key_last($cloud_keys);
     foreach ($local_keys as $local_index => $local_file) {
-      $last_key = array_key_last($cloud_keys);
       foreach ($cloud_keys as $index => $cloud_key) {
         if (trim($local_file->getContents()) === trim($cloud_key->public_key)) {
-            $sha256_hash = FingerprintGenerator::getFingerprint($cloud_key->public_key, 'sha256');
-            $table->addRow([
-              $cloud_key->label . PHP_EOL . $cloud_key->uuid,
-              $local_file->getFilename(),
-              $sha256_hash . PHP_EOL . $cloud_key->fingerprint,
-            ]);
-            if ($last_key !== $index) {
-              $table->addRow(new TableSeparator());
-            }
+          $sha256_hash = FingerprintGenerator::getFingerprint($cloud_key->public_key, 'sha256');
+          $table->addRow([
+            $cloud_key->label . PHP_EOL . $cloud_key->uuid,
+            $local_file->getFilename(),
+            $sha256_hash . PHP_EOL . $cloud_key->fingerprint,
+          ]);
+          if ($last_key !== $index) {
+            $table->addRow(new TableSeparator());
+          }
           unset($cloud_keys[$index], $local_keys[$local_index]);
           break;
         }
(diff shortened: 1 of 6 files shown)


grasmash:
Thank you. The idea is good.

I will have to take a look at the output. Glancing at the code, it seems like we should be creating separate columns for the fingerprints and using a table separator rather than an EOL character

grasmash:
After looking at the output, I see why you concatenated values in the columns. But I don't love doing that.

Anyway, I pushed up a commit to add table separators and some whitespace to make it a bit easier to understand.

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -33,21 +35,57 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     $local_keys = $this->findLocalSshKeys();
 
     $table = new Table($output);
-    $table->setHeaders(['Local Key Filename', 'Cloud Platform Key Label']);
+    $table->setHeaders([
+      'Cloud platform key label and UUID',
```
anavarre, line 39: `platform` should take a capital `P`

Also, I don't have a better suggestion, but I don't really like the label and UUID side-by-side. It lacks readability.

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -33,21 +35,57 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+      'Local key filename',
+      'Hashes — sha256 and md5',
```
anavarre, line 41: Any reason why we need both hashes?

tests/phpunit/src/Commands/Ssh/SshKeyListCommandTest.php
```diff
@@ -37,8 +37,8 @@ public function testUpload(): void {
     // Assert.
     $this->prophet->checkPredictions();
     $output = $this->getDisplay();
-    $this->assertStringContainsString('Local Key Filename', $output);
-    $this->assertStringContainsString('Cloud Platform Key Label', $output);
+    $this->assertStringContainsString('Local key filename', $output);
+    $this->assertStringContainsString('Cloud platform key label', $output);
```
anavarre, line 41: Capital `P` on `platform`.

anavarre:
If we're concerned about display methods, why are we not going with a `ssh-key info` command instead? This way we'd have a way more readable view. E.g.

```
$ acli ssh-key:info myapp
Please select an SSH key to get more information:
  [0] demo
  [1] drupal9
  [2] umami
 > 0
 ---------------------------- -------------------------------------------------------------------------------- 
  SSH Key Property             SSH Key Value                                                                       
 ---------------------------- -------------------------------------------------------------------------------- 
  Local Key Filename           id_rsa_acquia_ide_e45a3883-b4b1-4174-9aed-5ab5978a37fe.pub
  Cloud Platform Key Label     IDE_demo_ee76ad47b5ef4678858271fa217f95c1
  Cloud Platform Key UUID      a8ba0b7b-3daa-4824-9bcb-66200dae058d
  sha256 hash                  8Rwd3eGOERrxZ75ZSmY89d0odd9vmcWOSnULoL78cm0=
  md5 hash                     4c:db:fc:59:32:62:32:9e:7c:62:f2:33:68:1f:70:1d                                                       
 ---------------------------- -------------------------------------------------------------------------------- 
```

src/Command/Ssh/SshKeyListCommand.php
grasmash: Yeah I'm wondering if we should change the display so that it prints sections rather than a table. E.g.,

```
-------------------
Label:
UUID:
Filename:
sha256:
md5:
-------------------
Label:
UUID:
Filename:
sha256:
md5:
-------------------
```

grasmash: Just did some prototyping, would be:

```
Cloud keys with matching local keys
===================================

 ---------------- -------------------------------------------------
  Label            id_rsa_ac
  UUID             ...
  Local filename   id_rsa.pub
  sha256 hash      ...
  md5              ...
 ---------------- -------------------------------------------------

 ---------------- -------------------------------------------------
  Label            MattGrasmickMBP
  UUID             ...
  Local filename   acquia_cloud.pub
  sha256 hash      ...
  md5              ...
 ---------------- -------------------------------------------------

Cloud keys with no matching local keys
======================================

 ---------------- -------------------------------------------------
  Label            PT Travis CI
  UUID             ...
  Local filename   none
  sha256 hash      ...
  md5              ...
 ---------------- -------------------------------------------------
```

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
 
-    $table = new Table($output);
-    $table->setHeaders([
-      'Cloud platform key label and UUID',
-      'Local key filename',
-      'Hashes — sha256 and md5',
-    ]);
-    // First list all keys for which there are both Cloud and local keys.
-    $last_key = array_key_last($cloud_keys);
+    $this->io->title('Cloud keys with matching local keys');
```
anavarre, line 37: "Cloud Platform"

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-            $sha256_hash . PHP_EOL . $cloud_key->fingerprint,
-          ]);
-          if ($last_key !== $index) {
-            $table->addRow(new TableSeparator());
-          }
+          $this->io->definitionList(
+            ['Label' => $cloud_key->label],
+            ['UUID' => $cloud_key->uuid],
+            ['Local filename' => $local_file->getFilename()],
+            ['sha256 hash' => $sha256_hash],
```
anavarre, line 46: If we had the string "hash" here we probably need it for the `md5` hash below.

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+            ['md5' => $cloud_key->fingerprint]
+          );
           unset($cloud_keys[$index], $local_keys[$local_index]);
           break;
         }
       }
     }
-    // Second list all cloud keys for which there is no local key.
-    $last_key = array_key_last($cloud_keys);
+    $this->io->title('Cloud keys with no matching local keys');
```
anavarre, line 54: "Cloud Platform"

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-      ]);
-      if ($last_key !== $index || $local_keys) {
-        $table->addRow(new TableSeparator());
-      }
+      $this->io->definitionList(
+        ['Label' => $cloud_key->label],
+        ['UUID' => $cloud_key->uuid],
+        ['Local filename' => 'none'],
+        ['sha256 hash' => $sha256_hash],
+        ['md5' => $cloud_key->fingerprint]
```
anavarre, line 62: Same "hash" string missing here.

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+      );
     }
 
-    // Last list all local keys for which there is no cloud key.
-    $last_key = array_key_last($local_keys);
+    $this->io->title('Local keys with no matching cloud keys');
```
anavarre, line 66: s/cloud/Cloud Platform/g

src/Command/Ssh/SshKeyListCommand.php
```diff
@@ -34,60 +34,47 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
-      ]);
-      if ($last_key !== $index) {
-        $table->addRow(new TableSeparator());
-      }
+      $this->io->definitionList(
+        ['Label' => 'none'],
+        ['UUID' => 'none'],
+        ['Local filename' => $local_file->getFilename()],
+        ['sha256 hash' => $sha256_hash],
+        ['md5' => $md5_hash]
```
anavarre, line 75: Ditto.

src/Command/Ssh/SshKeyListCommand.php
grasmash: @anavarre what do you think about this output format?

anavarre: I was more in favor of keeping the current command simple and having an `ssh-key:info` command specifically. If you want to keep it all with one command, then I guess that works.

grasmash:
@eporama Would satisfy your needs for us to introduce an ssh-key:info command that shows all the data for a single key at a time?

danepowell:
If we can do it cleanly, then showing everything (including the fingerprint hash) in ssh-key:list is probably better than having a separate command. Though I don't have a very strong opinion as I don't use it often myself. If the output is ungainly, maybe we can at least show truncated hashes in ssh-key:list and full hashes in ssh-key:info?

Consider the following use case: you know the hash of your local SSH key and you want to see if it exists remotely. If ssh-key:list shows hashes, it's just one command and you can review the list to find your hash. With ssh-key:info, you have to check each key individually.

anavarre:
> Consider the following use case: you know the hash of your local SSH key and you want to see if it exists remotely. If ssh-key:list shows hashes, it's just one command and you can review the list to find your hash. With ssh-key:info, you have to check each key individually.

Yes, that's exactly what I had in mind and then realized it's something you almost never need, right?

tests/phpunit/src/Commands/Ssh/SshKeyListCommandTest.php
```diff
@@ -38,7 +38,10 @@ public function testUpload(): void {
     $this->prophet->checkPredictions();
     $output = $this->getDisplay();
     $this->assertStringContainsString('Local filename', $output);
-    $this->assertStringContainsString('Cloud Platform key label', $output);
+    $this->assertStringContainsString('Label', $output);
+    $this->assertStringContainsString('UUID', $output);
+    $this->assertStringContainsString('sha256', $output);
```
anavarre, line 43: `hash` or no `hash`, that's the question.

grasmash: This is just a partial string match to ensure the label is there.

PR #417, opened 2021-02-19
Use IO to ask questions and display default values.

grasmash:


src/Command/Ssh/SshKeyCreateCommand.php
```diff
@@ -88,12 +88,12 @@ protected function determineFilename(InputInterface $input, OutputInterface $out
       $this->validateFilename($filename);
     }
     else {
-      $question = new Question('<question>Please enter a filename for your new local SSH key:</question> ');
+      $question = new Question('<question>Please enter a filename for your new local SSH key:</question> ', 'id_rsa_acquia');
```
grasmash, line 91: Added default ssh key name.

anavarre: Do we have a check to ensure to not overwrite an existing key with the same name?

pratikpp: Should we also add a sequence number of the IDE - in case of multiple entitlements. THey could have different SSH keys

grasmash: No, you'd never have keys for multiple IDEs inside of one IDE. The label stored in Cloud does have such a suffix, but it's not required for a local filename.

src/Command/Ssh/SshKeyCreateUploadCommand.php
```diff
@@ -42,6 +42,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
       '--no-wait' => $input->getOption('no-wait'),
     ];
     $list_input = new ArrayInput($arguments);
+    $list_input->setStream($input->getStream());
```
grasmash, line 45: @danepowell This was necessary to pass the input stream set by phpunit tests.

danepowell (approved):
Looks good, but can you also delete the (now stale, I presume) `CommandBase->questionHelper` property?

PR #419, opened 2021-02-19
Resolve API token access issue.

grasmash:


webchick:
I was having this problem where no matter how many times I `acli auth:login`ed successfully, on subsequent commands it kept telling me:

```
  Your Cloud Platform API credentials are invalid. Run acli auth:login to reset them.
```

Manually applied this patch and ran `acli ckc` (per Matthew) and I'm back in business! Thank you! :)

danepowell (approved):

PR #420, opened 2021-02-22
Fix #398: Change environment id, application id, and directory to options

### LINKED ISSUE #398: acli pull environmentId option mistakenly believes the user is in an unexpected directory
```
ide:~/project (master) $ acli pull environmentId=REDACTED_ENV_UUID

In PullCommandBase.php line 132:
                                                                
  Please run this command from the /home/ide/project directory  
                                                                
```

danepowell:
**Motivation**
Fixes #398

**Proposed changes**
Change the `environmentId`, `applicationId`, and `dir` arguments to options, since they are all technically optional and it can be confusing when you want to provide one without the other (see #398 for example). Options are more verbose, but also more functional. They are optional because if empty, we (almost?) always prompt to select the environment/application.

A few examples of how commands would change:

- `acli pull . myapp.dev` becomes `acli pull --dir . --environmentId myapp.dev`
- `acli log:tail myapp.dev` becomes `acli log:tail --environmentId myapp.dev`

Note that this is a rather seriously breaking change, so we'd ideally cut a new major release, and we need to tell people that these arguments are changing to parameters.

**Alternatives considered**
I don't have any other ideas about how to solve #398. There's no way to provide one argument without providing all arguments.

**Testing steps**

3. Test any commands that use environment and application arguments, especially API and pull commands.

**Merge requirements**

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -385,20 +385,18 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
    * Add argument and usage examples for applicationUuid.
    */
   protected function acceptApplicationUuid() {
-    $this->addArgument('applicationUuid', InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias.')
-    ->addUsage(self::getDefaultName() . ' [<applicationAlias>]')
-    ->addUsage(self::getDefaultName() . ' myapp')
-    ->addUsage(self::getDefaultName() . ' abcd1234-1111-2222-3333-0e02b2c3d470');
+    return $this->addOption('applicationUuid', NULL, InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias.')
+    ->addUsage(self::getDefaultName() . ' --applicationUuid myapp')
+    ->addUsage(self::getDefaultName() . ' --applicationUuid abcd1234-1111-2222-3333-0e02b2c3d470');
   }
 
   /**
    * Add argument and usage examples for environmentId.
    */
   protected function acceptEnvironmentId() {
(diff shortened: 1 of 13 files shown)


danepowell:
I'm aware that tests are broken, but I want preliminary approval on the approach before proceeding since there's a lot of internal logic depending on these arguments.

grasmash:
I would suggest making the change smaller. We will resolve the user experience issue if we limit the number of possible arguments to one. I think that passing environment and application id as optional arguments is a better user experience. Therefore, we can simply change the directory argument to a parameter and I think that the problem is solved.

anavarre:
The reason why I filed #398 is only because in an IDE I'd get this:

```
ide:~/project (master) $ acli pull environmentId=REDACTED_ENV_UUID

In PullCommandBase.php line 132:
                                                                
  Please run this command from the /home/ide/project directory                                                              
```

I didn't know doing this would actually force me into passing `acli pull . environmentId=REDACTED_ENV_UUID`. Having the target directory abstracted out would have satisfied the requirement in an IDE. I understand we might still want to improve the UX outside of an IDE, though.

Now, if we want to go about changing those arguments to options, I'd like to suggest we make sure to not degrade the UX. I feel the below is more complicated and not a UX win for end users:

> acli log:tail myapp.dev becomes acli log:tail --environmentId myapp.dev

grasmash:
So I'd propose this scope: change the `dir` argument to an option. That's all :)

danepowell:
Thanks, I figured putting up the wrong solution might encourage consensus on the right one 😄 https://github.com/acquia/cli/pull/422

PR #423, opened 2021-02-23
Support legacy API keys.

grasmash:
Currently, key's created in ACLI < 1.6.0 are not properly inserted into the services container. This PR resolves that issue.


anavarre:
Do we know which minimal version can be considered legacy?

grasmash:
< 1.6.0.

anavarre:
Wow.

grasmash:
We have 2 customer confirmations that this fixes the bug.

PR #429, opened 2021-02-24
Use consistent help text for environmentId.

grasmash:


src/Command/Pull/PullCodeCommand.php
```diff
@@ -21,10 +21,10 @@ class PullCodeCommand extends PullCommandBase {
   protected function configure() {
     $this->setDescription('Copy code from a Cloud Platform environment')
       ->addArgument('dir', InputArgument::OPTIONAL, 'The directory containing the Drupal project to be refreshed')
-      ->addArgument('environmentId', InputArgument::OPTIONAL, 'The UUID of the associated Cloud Platform source environment')
       ->addOption('no-scripts', NULL, InputOption::VALUE_NONE,
         'Do not run any additional scripts after code is pulled. E.g., composer install , drush cache-rebuild, etc.')
       ->setHidden(!AcquiaDrupalEnvironmentDetector::isAhIdeEnv() && !self::isLandoEnv());
+    $this->acceptEnvironmentId();
```
danepowell, line 27: @grasmash you could make `acceptEnvironmentId()` return the output of `addArgument()`, and then you can chain it like every other `addArgument()` call here.

grasmash: Good idea.

danepowell (approved):

PR #430, opened 2021-02-24
Fixes #425: acli new formatting is off in the Cloud IDE terminal.

### LINKED ISSUE #425: acli new formatting is off in the Cloud IDE terminal
The druplicon is onto the Drupal string:

acli-formatting-off

grasmash:


src/Command/NewCommand.php
```diff
@@ -66,7 +66,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     $this->initializeGitRepository($dir);
 
     $output->writeln('');
-    $output->writeln("<info>New 💧Drupal project created in $dir. 🎉");
+    $output->writeln("<info>New 💧 Drupal project created in $dir. 🎉");
```
anavarre, line 69: That works, but it seems we also have an extra white space after "New" FYI
Screenshot_20210225_154716

grasmash: It's not extra. Just 1 space.

PR #432, opened 2021-02-25
Switch from bug/feature to major/minor/patch labels

danepowell:
**Motivation**
I find our adherence to semver a little stifling. Linking bug fixes/features to minor/patch versions doesn't reflect (even in the slightest) the potential risk/impact/value to a customer. A huge bug fix like https://github.com/acquia/cli/pull/423 warrants more attention than a patch release. Conversely, a tiny enhancement like https://github.com/acquia/cli/pull/429 barely warrants a patch release. Yet we are stuck with bug fix == patch release and feature == minor release because of semver.

Especially in releases with multiple issues, I think the biggest changes should be featured most prominently, rather than grouping changes by bug fix/feature.

Removing a layer of abstraction and controlling the version number directly seems easiest here.

**Proposed changes**
Label PRs as major/minor/patch changes, and use these to determine release version numbers.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,5 +17,5 @@ How can we replicate the issue and verify that this PR fixes it?
 3. (add specific steps for this pr)
 
 **Merge requirements**
-- [ ] _Bug_, _enhancement_, or _breaking change_ label applied
+- [ ] _Major_, _Minor_, or _Patch_ label applied
 - [ ] Manual testing by a reviewer
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -2,25 +2,23 @@ name-template: '$RESOLVED_VERSION'
 tag-template: '$RESOLVED_VERSION'
 categories:
   - title: '🚨 Major changes'
-    label: 'breaking change'
-  - title: '🚀 Enhancements'
-    label: 'enhancement'
-  - title: '🐛 Bug Fixes'
-    label: 'bug'
-  - title: '🧰 Maintenance'
-    label: 'chore'
+    label: 'major'
+  - title: '🚀 Minor changes'
+    label: 'minor'
+  - title: '🧰 Patch changes'
+    label: 'patch'
 change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
 change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
 version-resolver:
   major:
     labels:
-      - 'breaking change'
+      - 'major'
   minor:
     labels:
-      - 'enhancement'
+      - 'minor'
   patch:
     labels:
-      - 'bug'
+      - 'patch'
   default: patch
 template: |
   ## What's new since $PREVIOUS_TAG


anavarre:
I understand, but then we are losing the release drafter's primary value which is to understand what is a bug fix and what is a feature. Why wouldn't we simply add more labels, especially to warn about a potentially breaking change?

danepowell:
> release drafter's primary value which is to understand what is a bug fix and what is a feature

Do end users really care that much about whether something is a bug fix, feature, chore, or any other category we come up with? Don't they care more about how a change impacts them? For instance, "we changed every command name" is a huge change that people should pay attention to. "we fixed a typo" is hardly worth mentioning. Now tell me, are either of those a bug or a feature? Does it matter?

anavarre:
End users, no. Developers, yes. Acquia CLI is first and foremost aimed at developers so I think they would care.

See also https://www.conventionalcommits.org/en/v1.0.0-beta.4/#commit-message-with-description-and-breaking-change-in-body

danepowell:
Quite the coincidence that semver (tying version numbers to enhancements/bug fixes rather than issue severity/impact) bit us twice this morning:
- In https://github.com/acquia/cli/issues/437, Composer issued a new patch release with a breaking change. It was a patch release because technically it was a "bug fix" so no one paid attention, and it broke our builds. But regardless of the explicit spec, the _implicit_ meaning of a patch release is something trivial and non-breaking. This discrepancy between the explicit and implicit usage of semver is what bit us.
- In fixing the broken phar, we want to push out 1.7.2 (since the intervening changes are relatively minor) but because technically they are enhancements, the next release has to be 1.8.0. Which implies that it's a more significant release than it really is (a hotfix should be a patch release).

anavarre:
We need to decide whether we want to adhere to this https://semver.org/spec/v2.0.0.html#summary

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> * MAJOR version when you make incompatible API changes,
> * MINOR version when you add functionality in a backwards compatible manner, and
> * PATCH version when you make backwards compatible bug fixes.
>
> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

danepowell:
I'm saying semver essentially doesn't apply to us, and we shouldn't follow it slavishly. We should still follow it in _spirit_, just not in letter. From the first paragraph of the very same semver doc:

> For this system to work, you first need to declare a public API.

That's essentially the root of the problem. ACLI is a command-line tool, not a library. We expose no public services or API. The only "downstream consumers" are humans, and they need to be treated differently than machines.

I'll change this PR to use a hybrid system like I proposed, so that release notes can still be organized by bug fix / enhancement, hopefully that's acceptable to everyone.

anavarre:
> I'll change this PR to use a hybrid system like I proposed, so that release notes can still be organized by bug fix / enhancement, hopefully that's acceptable to everyone.

Sounds good, but you might have to hold my hand when trying to aim for a new release when I'm trying to jump to a new minor when you want a patch release.

danepowell:
Yeah on second thought, I think I'm just going to put this down and step away. Having minor/patch tags begs the question of how we train people to categorize things as minor/patch, and I have no interest in going down that road. I still feel like the current system is broken, but I'm not clear on how to make it less broken.

PR #442, opened 2021-02-26
Fixes #424: Trying to get property sshUrl of non-object.

### LINKED ISSUE #424: Notice: Trying to get property 'sshUrl' of non-object in phar://opt/local/acli/src/Helpers/SshHelper.php on line 158
User ran into the following error message when clicking on the IDE "Configure IDE" button in a brand new IDE

```
Notice: Trying to get property 'sshUrl' of non-object in phar://opt/local/acli/src/Helpers/SshHelper.php on line 158
```

Never ran into this. This is happening with 1.5.0 (asked the user to update to 1.7.0 and report back)

grasmash:


src/Command/Ide/Wizard/IdeWizardCreateSshKeyCommand.php
```diff
@@ -237,12 +237,15 @@ protected function userHasUploadedIdeKeyToCloud(): bool {
...
    * @throws \Exception
    */
-  protected function getDevEnvironment($cloud_app_uuid): ?EnvironmentResponse {
+  protected function getAnyAhEnvironment(string $cloud_app_uuid): ?EnvironmentResponse {
     $acquia_cloud_client = $this->cloudApiClientService->getClient();
     $environment_resource = new Environments($acquia_cloud_client);
     $application_environments = iterator_to_array($environment_resource->getAll($cloud_app_uuid));
     foreach ($application_environments as $environment) {
-      if (AcquiaDrupalEnvironmentDetector::isAhDevEnv($environment->name)) {
+      if (AcquiaDrupalEnvironmentDetector::isAhDevEnv($environment->name)
```
danepowell, line 245: @grasmash do we even care if it's dev/stage/prod? Should we just return the first environment found?

anavarre: That's a good point. E.g. when we have Cloud IDE for ACF they won't have a prod env unless they upgrade to ACP.

grasmash: I think that production is the only environment that has a different set of permissions, so perhaps we should only check that it’s not production.

danepowell: > perhaps we should only check that it’s not production

Agreed. Let's also throw an exception in the event no suitable environment is found, or gracefully bail on the SSH check but finish the rest of the command. Anything except a PHP error and silent failure.

danepowell (approved):

PR #444, opened 2021-02-26
Fixes #434: Override the spec to better set expectations when an IDE is going to be deleted.

### LINKED ISSUE #434: Override the spec to better set expectations when an IDE is going to be deleted
Right now, when deleting an IDE, we return the deletion message from the spec:

> The Cloud IDE is being deleted.

It gets users confused, because K8s takes a few minutes before actually processing the deletion. We should override the message and better set expectations. Here's a suggestion:

> The Cloud IDE will be deleted momentarily. This process usually takes a few minutes.

grasmash:
Fix #434


anavarre:
I couldn't get this to work against `master` just now. Still getting `The Cloud IDE is being deleted.`.

PR #447, opened 2021-03-01
Fixes #436: Suggest a default IDE label to not let the new user confused as to what is expected.

### LINKED ISSUE #436: Suggest a default IDE label to not let the new user confused as to what is expected
Passing a label upon creating an IDE can be confusing for the new user (they don't always understand it's arbitrary). We should likely offer to have a default name so the user can just press enter.

Something like First Last would work, e.g.

```
acli api:accounts:find | jq '(.first_name + " " + .last_name)' | tr -d \"
```

While we're at it, we could probably improve the string when asking the user to pass a label. Maybe make it more personal?

> Which label would you like to give to your IDE?

grasmash:


tests/phpunit/src/Commands/Ide/IdeCreateCommandTest.php
```diff
@@ -64,8 +70,7 @@ public function testCreate(): void {
     $this->assertStringContainsString('Please select a Cloud Platform application:', $output);
     $this->assertStringContainsString('  [0] Sample application 1', $output);
     $this->assertStringContainsString('  [1] Sample application 2', $output);
-    $this->assertStringContainsString('Please enter a label for your Cloud IDE:', $output);
-    // $this->assertStringContainsString('Waiting for DNS to propagate...', $output);
```
anavarre, line 68: Why are we commenting this out?

anavarre (approved):
Creating:

```
$ acli ide:create myapp

 Please enter a label for your Cloud IDE [John Doe's IDE]:
 > 
```

Listing:

```
$ acli ide:list myapp
 ======================================================================= 
  IDEs                                                                   
 ======================================================================= 
  John Doe's IDE (john@doe.com)                   
  IDE URL: https://f4f3c839-4226-49f6-a374-4e5a6a2285bd.ide.ahdev.cloud  
  Web URL: https://f4f3c839-4226-49f6-a374-4e5a6a2285bd.web.ahdev.cloud  
 ===================================================================
```

tests/phpunit/src/Commands/Ide/IdeCreateCommandTest.php
grasmash: I removed the comment here.

PR #449, opened 2021-03-01
Fixes #435: Users are confused with api sub-commands.

### LINKED ISSUE #435: Users are confused with api sub-commands placed halfway through the acli output and too many commands in general
Right now, when typing `acli`, users miss that we have way more API commands available, despite the cyan color. This is because the below is shown ~halfway through the output:

```
 api
  api:accounts:find                        Returns details about your account.
  api:environments:code-deploy             Deploys code to this environment.
  api:environments:database-backup-create  Create a backup.
  api:environments:domain-clear-varnish    Clears the Varnish cache for the specified domain.
  api:list                                 [api] There are more hidden API commands! Run api:list to see them all.
```

It's also confusing to them what the `api` grouping really means and they find we have many commands on screen.

-----

We could explore several improvements:

1. Place `api:*` sub-commands at the very end of the `acli` output so they'd stand out immediately (but then we'd lose the A-Z ordering)
2. Only show `api:list` with a even more descriptive message, without suggesting common commands we assume would be most helpful for a large fraction of our users

Perhaps it's a "not a bug, won't fix, but let's explore whether there's anything we can think of.

grasmash:


grasmash:
image

src/Command/Api/ApiListCommand.php
```diff
@@ -10,34 +10,33 @@
...
+class ApiListCommand extends ApiListCommandBase {
 
   protected static $defaultName = 'api:list';
 
   /**
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription("<fg=cyan>There are more hidden API commands! Run api:list to see them all.</>")
+    $this->setDescription("List all API commands")
```
anavarre, line 21: Do we not need to keep the color here to have it stand out?

grasmash: I don't think so. See uploaded image.

anavarre: I saw, thus why I was asking.

grasmash: I feel that the api:list command is less important now that we have a number of list commands. So, it doesn't seem necessary for it to stand out. I think we'd prefer the user to run one of the commands with less output.

anavarre (approved):
From the UX standpoint, the PR works as intended.

anavarre (changes requested):
There's one issue I could spot related to a sub-string partial match or equivalent. IDE commands return:

```
Available commands for the "api" namespace:
  api:ide:find                    Returns Cloud IDE info.
  api:identity-providers          List all API commands for the identity-providers resource
  api:identity-providers:delete   Deletes a specific identity provider by its UUID.
  api:identity-providers:disable  Disables an identity provider by its UUID.
  api:identity-providers:enable   Enables an identity provider by its UUID.
  api:identity-providers:find     Returns a specific identity provider by UUID.
  api:identity-providers:list     Returns a list of identity providers for a user.
  api:identity-providers:update   Modifies an identity provider by its UUID.
```

`ide` and `identity` are treated the same.

grasmash:
Addressed in the latest commit.

anavarre (approved):

danepowell (approved):

PR #451, opened 2021-03-01
Fixes #450: Add contextual information to exceptions.

### LINKED ISSUE #450: Add contextual information to exceptions
E.g., we should link to Acquia CLI docs and other relevant resources.

grasmash:


bin/acli
```diff
@@ -17,6 +17,7 @@
 namespace Acquia\Cli;
 
 use Acquia\Cli\Command\Api\ApiCommandHelper;
+use Acquia\Cli\EventListener\ExceptionListener;
```
danepowell, line 20: This include shouldn't be necessary, should it?

config/dev/services.yml
```diff
@@ -0,0 +1,12 @@
+imports:
+  - { resource: '../prod/services.yml' }
+
+services:
```
danepowell, line 4: I'm not sure I understand the purpose of this bifurcation. It seems like you're inheriting a default set of prod services, and then overriding them for dev?

I think we build the phar in production mode, so if you need BufferedOutput or something else from this file for formatting to work, it won't work with the phar.

src/Command/CommandBase.php
```diff
@@ -1217,16 +1217,20 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
...
       } catch (ValidatorException $validator_exception) {
         // Since this isn't a valid UUID, let's see if it's a valid alias.
         $alias = $this->normalizeAlias($application_uuid_argument);
-        $customer_application = $this->getApplicationFromAlias($alias);
-        $input->setArgument('applicationUuid', $customer_application->uuid);
+        try {
+          $customer_application = $this->getApplicationFromAlias($alias);
+          $input->setArgument('applicationUuid', $customer_application->uuid);
+        } catch (AcquiaCliException $exception) {
+          throw new AcquiaCliException("{applicationUuid} must be a valid UUID or application alias.");
```
danepowell, line 1232: Is there any other reason that the enclosed code would throw an exception? Like if auth fails or there's a network error, will people get a false positive about the UUID being invalid?

I'd be a little more comfortable if we check the caught exception message before rethrowing.

danepowell (changes requested):

config/dev/services.yml
grasmash: This is only used during testing. It’s necessary because during testing, we used buffered output instead of the console output. The same issue occurs for input.

Without this, services are auto wired incorrectly during testing.

bin/acli
grasmash: Correct, this is an artifact from a previous iteration.

src/Command/CommandBase.php
grasmash: Notice that we are only catching exceptions of type AcquiaCliException. Other exceptions will not be caught.

config/dev/services.yml
grasmash: > so if you need BufferedOutput or something else from this file for formatting to work, it won't work with the phar.

BufferedOutput is not required for formatting. It's required for capturing output during testing.

grasmash:
Rebased here: https://github.com/acquia/cli/pull/452

PR #452, opened 2021-03-02
Improve contextual help in command output

grasmash:


src/EventListener/ExceptionListener.php
```diff
@@ -7,40 +7,108 @@
...
+  protected function writeApplicationAliasHelp(SymfonyStyle $io): void {
+    $io->comment('<options=bold>applicationUuid</> can also be an application alias. E.g. <options=bold>myapp</>.' . PHP_EOL
+    . 'Run <options=bold>acli remote:aliases:list</> to see a list of all available aliases.');
+  }
+
+  /**
+   * @param \Symfony\Component\Console\Style\SymfonyStyle $io
+   */
+  protected function writeSiteAliasHelp(SymfonyStyle $io): void {
+    $io->comment('<options=bold>environmentId</> can also be an site alias. E.g. <options=bold>myapp.dev</>.' . PHP_EOL
```
anavarre, line 103: `a` site alias?

src/EventListener/ExceptionListener.php
```diff
@@ -7,40 +7,108 @@
...
         default:
-          $newErrorMessage = 'Cloud Platform API returned an error: ' . $errorMessage;
+          $new_error_message = 'Cloud Platform API returned an error: ' . $errorMessage;
       }
+      $io->comment("You can learn more about Cloud Platform API at <href=https://docs.acquia.com/cloud-platform/develop/api/>docs.acquia.com</>");
     }
 
-    if (isset($newErrorMessage)) {
-      $event->setError(new AcquiaCliException($newErrorMessage, [], $exitCode));
+    $io->comment("You can find Acquia CLI documentation at <href=https://docs.acquia.com/acquia-cli/>docs.acquia.com</>");
```
anavarre, line 85: Are you not missing surrounding quotes for the hyperlink? It's not clickable for me here.

src/EventListener/ExceptionListener.php
```diff
@@ -7,40 +7,108 @@
...
         case "There are no available Cloud IDEs for this application.\n":
-          $newErrorMessage = $errorMessage . "Delete an existing IDE (acli ide:delete) or contact your Account Manager or Acquia Sales to purchase additional IDEs.\n You may also submit a support ticket to ask for more information (https://insight.acquia.com/support/tickets/new?product=p:ride)";
+          $io->comment("Delete an existing IDE via <options=bold>acli ide:delete</> or contact your Account Manager or Acquia Sales to purchase additional IDEs.");
+          $this->writeSupportTicketHelp($io);
           break;
         default:
-          $newErrorMessage = 'Cloud Platform API returned an error: ' . $errorMessage;
+          $new_error_message = 'Cloud Platform API returned an error: ' . $errorMessage;
       }
+      $io->comment("You can learn more about Cloud Platform API at <href=https://docs.acquia.com/cloud-platform/develop/api/>docs.acquia.com</>");
```
anavarre, line 82: Not clickable.

config/dev/services.yml
```diff
@@ -0,0 +1,12 @@
+imports:
+  - { resource: '../prod/services.yml' }
+
+services:
+  Symfony\Component\Console\Output\BufferedOutput: ~
```
grasmash, line 5: This is only used during testing. It’s necessary because during testing, we use BufferedOutput instead of the ConsoleOutput. The same issue occurs for input--we need to switch from ArgvInput to ArrayInput.

Without this services file override, services will be auto wired incorrectly during testing.  We would end up with multiple inputs and outputs of different types. For example, the production services file would create an ArgvInput and a ConsoleOutput for all services accepting InputInterface and OutputInterface arguments, like the ExceptionListener and the TelemetryHelper.

Our test would then create a separate ArrayInput and BufferedOutput and inject those into the Application. At that point the application would be using a different input and output than the other services in the container. This causes all sorts of problems, not least of which is our inability to capture and assert the value of the output of services.

With the currently proposed approach, we correctly inject these services into the container and obviate the need to create them manually. We also ensure that there is only one input and one output object and that those objects are shared across services.

anavarre:
Overall it's a good improvement. I think we could improve theming some more.

```
 Please enter a label for your Cloud IDE [Aurelien's IDE]:
 > 

    ⌛ Creating your Cloud IDE...

 // Delete an existing IDE via acli ide:delete or contact your Account Manager or Acquia Sales to                       
 // purchase additional IDEs.                                                                                           

 // You may also submit a support ticket to ask                                                                         
 // for more information.                                                                                               

 // You can learn more about Cloud Platform API at                                                                      
 // docs.acquia.com                                                                                                     

 // You can find Acquia CLI documentation at docs.acquia.com                                                            


In Client.php line 154:
                                                           
  There are no available Cloud IDEs for this application.  
                                                           

ide:create [<applicationUuid>]

```

Instead of having prefixing `//` we could have those strings be all added to a grey-ish background color so the info block would stand out compared to the error message only. I also feel we're returning to a new line at ~45 chars and there's no obvious reason why we should be doing this. Can't the text auto-wrap when needed, e.g. in case you resize your terminal window?

config/dev/services.yml
grasmash: I've actually been trying to figure out how to do this for a while. I made an attempt in https://github.com/acquia/cli/pull/326, but it was really just a workaround. We still need to refactor TestBase: https://github.com/acquia/cli/blob/8d03cc72d75aa127a8c7fa82cb2936edb506403b/tests/phpunit/src/TestBase.php#L224-L231

The container should autowire itself during testing, this type of manually input and outputting per service isn't quite right.

danepowell (approved):

PR #455, opened 2021-03-03
Fixes #433: Infinite memory usage results in acli login

### LINKED ISSUE #433: Infinite memory usage results in kill during `acli auth:login -n`
I'm implementing Acquia integration into [ddev](https://github.com/drud/ddev) in https://github.com/drud/ddev/pull/2837 and all was going well until tests started reporting failures during authentication (`acli auth:login -n --key="${ACQUIA_API_KEY}" --secret="${ACQUIA_API_SECRET}"`)

It turns out that acli uses up all available memory and then gets killed. 

This seems to have begin in 1.7.0, at least I can't recreate it in 1.6.0

This seems to be intermittent, but when it starts happening it happens a lot. 

The ddev integration is already in prerelease, see https://github.com/drud/ddev/releases/tag/v1.17.0-alpha5, which has instructions. This behavior appeared when I upgraded the docker image and picked up 1.7.0 AFAICT.

grasmash:


anavarre (approved):
Manual test

```
/ $ (issue-433-box) $ ack -A1 "box-install" ./composer.json
    "box-install": [
      "curl -O -L https://github.com/humbug/box/releases/latest/download/box.phar"
```

danepowell (approved):

PR #458, opened 2021-03-04
Updated build and release docs

danepowell:
- Add build steps to contributing doc
- Remove release doc, since release drafter manages everything now
- Move box.phar to build directory so it's more easily excluded from Git and IDE indexing

These steps are necessary to test the fix for https://github.com/consolidation/self-update/issues/9 via https://github.com/danepowell/self-update/pull/3

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -51,15 +51,22 @@ composer install
 
 Be sure to validate and test your code locally using the provided Composer test scripts (`composer test`) before opening a PR.
 
+### Building acli.phar
+
+To test changes in production mode, build and run acli.phar using this process. You can verify that the "Build PHAR" stage of .travis.yml follows a similar process.
+
+1. Install Composer production dependencies: `composer install --no-dev --optimize-autoloader`
+1. Clear and rebuild your Symfony cache: `./bin/acli ckc && ./bin/acli`
+1. Install Box and dump env vars (only need to do this once): `composer box-install && composer dump-env prod`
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -51,15 +51,41 @@ composer install
...
 Be sure to validate and test your code locally using the provided Composer test scripts (`composer test`) before opening a PR.
 
+### Building acli.phar
+
+To test changes in production mode, build and run acli.phar using this process. You can verify that the "Build PHAR" stage of .travis.yml follows a similar process.
+
+1. Install Composer production dependencies: `composer install --no-dev --optimize-autoloader`
+1. Clear and rebuild your Symfony cache: `./bin/acli ckc && ./bin/acli`
+1. Install Box and dump env vars (only need to do this once): `composer box-install && composer dump-env prod`
+1. Compile phar: `composer box-compile`
```
anavarre, line 61: I tried to follow those steps and ran into:

```
The package "beberlei/assert" requires the extension "intl". Enable it or install a polyfill.
```

Should we document this dependency?

danepowell: I think this dependency was introduced by https://github.com/acquia/cli/pull/455

The error message is pretty clear, I don't see what we gain by adding documentation. But if it's not clear to you then feel free to suggest a change.

anavarre: It's clear to me, I was just wondering whether we wanted to document it or not.

PR #462, opened 2021-03-05
Fixes #461: Ability to set DB credentials via environment variables.

### LINKED ISSUE #461: Ability to set DB credentials via environment variables
In the current Acquia CLI implementation I think the only thing really missing if the ability to set the DB credentials via env variables. There is a set Lando specific env variables for it that can be used but the formatting is not ideal.

Basically if Acquia CLI would look to generic env vars for each of the DB settings that would be a big step forward.

grasmash:


.gitignore
```diff
@@ -10,6 +10,7 @@ drupal
 .idea
 build
 *.bak
+*.phar
```
danepowell, line 13: What file are you needing to ignore here? There shouldn't be any phars in the root directory.

danepowell:
There seems to be a lot of boilerplate / duplication here, which doesn't make me super happy. But I'm not sure how to improve it.

Instead of all of the getting/setting, you could pass the credentials via dependency injection, but that doesn't reduce the duplication...

grasmash:
For me, it didn’t quite rise to the level of necessitating its own service. It’s still very straightforward. I’m trying to avoid premature optimization.

.gitignore
grasmash: I found these get generated during local testing of phar builds.

danepowell: An extra .gitignore line won't hurt anything, but it might indicate something wrong with your local, since phars should never exist outside of the `build` dir following https://github.com/acquia/cli/pull/458

danepowell (approved):

PR #470, opened 2021-03-08
#Fixes: 468: Users get confused because there's no visible output when entering their API secret.

grasmash:


anavarre:
Alternatively, if we can use *************** perhaps it's best.

grasmash:
Not sure how much we can really change here, we're relying on Symfony's internal input handling: https://symfony.com/doc/current/components/console/helpers/questionhelper.html#hiding-the-user-s-response

grasmash:
Yeah, I don't know what we can do about this: https://github.com/symfony/console/blob/5.x/Helper/QuestionHelper.php#L432

grasmash:
IMO, this is not worth the effort to change. We can either accept the change proposed here or un-hide the input.

anavarre:
I'd say let's go with your initial suggestion, then.

anavarre:
Is this the expected theming?

470-theming

I feel we should have consistency for questions.

grasmash:
Hmmm. Looks like we did this differently for a reason: Windows compatibility.

grasmash:
image

anavarre (approved):
Looks good. Like how it's consistent now. FYI the key icon looks great but it doesn't work on Linux.

PR #471, opened 2021-03-08
Fixes #469: Users are confused with default options when generating an SSH key or creating an IDE.

### LINKED ISSUE #469: Users are confused with default options when generating an SSH key or creating an IDE
When generating a new SSH key or creating a new IDE, we pass a default option so the user can simply press enter and continue with the suggestion.

Turns out it can be confusing to some users so we should likely improve the wording.

grasmash:


src/Command/Ssh/SshKeyCreateCommand.php
```diff
@@ -88,7 +88,7 @@ protected function determineFilename(InputInterface $input, OutputInterface $out
       $this->validateFilename($filename);
     }
     else {
-      $question = new Question('<question>Please enter a filename for your new local SSH key:</question> ', 'id_rsa_acquia');
+      $question = new Question('<question>Please enter a filename for your new local SSH key (default is id_rsa_acquia):</question> ', 'id_rsa_acquia');
```
anavarre, line 91: I feel we can still improve this. Suggestions:

* Keep "default" but surround the suggestion by double quotes
* Replace "default" by `Press enter to use "ide_rsa_acquia"`

BTW we also need to change this when creating a Cloud IDE.

grasmash:
This is another instance where we're really relying on Symfony internals and it's probably not worth it to try and override. See https://github.com/symfony/console/blob/92e1b7c4381c3c51e327b99c7ff067411e291783/Helper/SymfonyQuestionHelper.php#L59

anavarre (approved):
Manual tests worked well. I think considering the low-level changes required this is a very good solution to the problem we're trying to solve.

PR #476, opened 2021-03-12
Fixes #472: Acquia CLI tries to pull the files directory even when no files directory exists yet.

### LINKED ISSUE #472: Acquia CLI tries to pull the files directory even when no files directory exists yet
Found with user testing. Acquia CLI returns the following when no Drupal site is installed on the Cloud Platform:

```
Choose a site
> [0]
```

The user is forced to select `0` and then Acquia CLI fails with an error because there's no source files directory to copy from. We'd need to find a way to abort before we get to the error.

grasmash:


grasmash:
@anavarre I don't have an empty app to test this against.

PR #478, opened 2021-03-12
Fix #467: Spinners clear terminal output

### LINKED ISSUE #467: On macOS 10.15.7 Catalina, any operation involving the spinner erases any previous output
Observed live with user testing Acquia CLI 1.8.1. When the spinner kicks in, every output on screen is erased until the spinner is at the top of the screen. Happened when:

- Waiting for an SSH key to be uploaded to the Cloud Platform
- Waiting for the IDE to be provisioned

danepowell:
**Motivation**

Fixes #467

**Proposed changes**
Patch Acquia CLI to include https://github.com/symfony/symfony/pull/40450, until that's merged and we can update to a new Symfony release that includes the fix.

**Alternatives considered**
Wait until the Symfony release and just suffer the bug.

**Merge requirements**
- [x] Manual testing by a reviewer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
     "acquia/drupal-environment-detector": "^1.2.0",
     "composer/semver": "^3.2",
     "consolidation/self-update": "dev-allow-unstable as 1.1",
+    "cweagans/composer-patches": "^1.7",
     "doctrine/cache": "^1.10",
     "drupol/phposinfo": "^1.6.5",
     "kevinrob/guzzle-cache-middleware": "^3.3",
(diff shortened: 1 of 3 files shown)


grasmash (approved):

glensc:
@danepowell are you running `composer update` instead of `composer require vendor/package`, because the composer lock commit includes unrelated updates.

danepowell:
Yes, I run `composer update`. Maybe not strictly best practice from a dev process standpoint, but we have no other process for keeping dependencies up to date. We just added dependabot but that's only for security updates.

glensc:
@danepowell you should at least make two commits then:
- https://www.freshconsulting.com/atomic-commits/

PR #494, opened 2021-03-23
Ensure every prompt has a default response

danepowell:
**Motivation**
It's annoying to not have a default response to prompts, _especially_ when there's only a single available option.

**Proposed changes**
Ensure every prompt has a default response, even if it's just the first option in whatever array of choices we present.

**Alternatives considered**
Wrap the Symfony `choice()` method in a utility class/method that ensures that every choice has a default even if it's not explicitly set. Seems like overkill.

**Testing steps**
3. Try commands like `acli new` or `acli pull:db` that force a selection from a list, and ensure a default is set.

**Merge requirements**

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -1396,7 +1396,7 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
     foreach ($acsf_sites['sites'] as $domain => $acsf_site) {
       $choices[] = "{$acsf_site['name']} ($domain)";
     }
-    $choice = $this->io->choice('Choose a site', $choices);
+    $choice = $this->io->choice('Choose a site', $choices, $choices[0]);
     $key = array_search($choice, $choices, TRUE);
     $sites = array_values($acsf_sites['sites']);
     $site = $sites[$key];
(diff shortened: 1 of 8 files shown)


anavarre (approved):

anavarre:
Manual tests.

With `acli pull:db` pressing enter worked.

```
 Choose a database [myapp (default)]:
  [0] myapp (default)
  [1] drupal9
 > 

    ⌛ Downloading Drupal database copy from the Cloud Platform...
    ✔ Downloading Drupal database copy from the Cloud Platform
    ✔ Importing Drupal database download
    ✔ Clearing Drupal caches via Drush
```

With `acli new`, pressing enter worked.

```
 Choose a starting project [acquia/drupal-recommended-project]:
  [0] acquia/drupal-recommended-project
  [1] acquia/drupal-minimal-project
```

src/Command/CommandBase.php
```diff
@@ -1396,7 +1397,7 @@ protected function promptChooseAcsfSite($cloud_environment) {
     foreach ($acsf_sites['sites'] as $domain => $acsf_site) {
       $choices[] = "{$acsf_site['name']} ($domain)";
     }
-    $choice = $this->io->choice('Choose a site', $choices);
+    $choice = $this->io->choice('Choose a site', $choices, $choices[0]);
```
grasmash, line 1400: I might check to see if there's only one choice and just skip asking if that's the case.

danepowell: I thought about that but figured we'd leave it for later, because it's not trivial to get right. We'd want to make it really clear that only a single choice was available, and we wouldn't want to do it for longer or dangerous operation (i.e. `ide:delete`).

PR #498, opened 2021-03-26
Fix #496: Deploy builds failing

### LINKED ISSUE #496: Deploy builds failing
Our 1.9.0 release failed to attach acli.phar and I had to do it manually: https://travis-ci.com/github/acquia/cli/jobs/494009242

In the raw log output, `php build/box.phar compile` was just returning `Not Found`. Which turns out to not be an error, but the literal contents of `build/box.phar`, because Box didn't attach box.phar to _their_ latest release: https://github.com/box-project/box/issues/530

We should consider pinning the Box release again (with a comment explaining why), and/or verify the integrity of the box.phar download so at least it doesn't fail silently. If this can be done as a one-liner, contribute it to the box.phar docs.

danepowell:
**Motivation**
Fixes #496

**Proposed changes**
Pin the Box version and add the curl `--fail` flag so the script will exit and alert us next time instead of failing silently. Also update the namespace (apparently the project changed ownership).

**Alternatives considered**
Switch to something other than Box that can be properly verified. These have been known issues for a while with no maintainer response:
- https://github.com/box-project/box/issues/487
- https://github.com/box-project/box/issues/530

**Testing steps**
1. See that the first commit fails (as it should, since Box master is broken)
2. See that the second commit passes (after pinning Box to known good release)

**Merge requirements**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
       "rm -rf cx-api-spec"
     ],
     "box-install": [
-      "curl -L https://github.com/humbug/box/releases/latest/download/box.phar -o build/box.phar"
+      "curl -f -L https://github.com/humbug/box/releases/latest/download/box.phar -o build/box.phar"
     ],
     "box-compile": [
       "php build/box.phar compile"
--- a/composer.json
+++ b/composer.json
@@ -106,7 +106,7 @@
       "rm -rf cx-api-spec"
     ],
     "box-install": [
-      "curl -f -L https://github.com/humbug/box/releases/latest/download/box.phar -o build/box.phar"
+      "curl -f -L https://github.com/box-project/box/releases/download/3.11.1/box.phar -o build/box.phar"
     ],
     "box-compile": [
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -106,7 +106,7 @@
       "rm -rf cx-api-spec"
     ],
     "box-install": [
-      "curl -L https://github.com/humbug/box/releases/latest/download/box.phar -o build/box.phar"
+      "curl -f -L https://github.com/box-project/box/releases/download/3.11.1/box.phar -o build/box.phar"
```
szepeviktor, line 109:
```suggestion
      "wget -O build/box.phar https://github.com/box-project/box/releases/download/3.11.1/box.phar"
```

Wget will set exist status to 8 when it is not an HTTP/404.

danepowell: Thanks! But I think we'll stick with curl since `-f` should have the same effect, and we know it's available everywhere in our infrastructure.

PR #500, opened 2021-03-29
Unpatch symfony console

danepowell:
Symfony 5.2.6 contains the changes that we were previously applying via this patch.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "react/event-loop": "^1.1",
     "symfony/cache": "^5.1",
     "symfony/config": "^5.1",
-    "symfony/console": "^5.1",
+    "symfony/console": "^5.2.6",
     "symfony/dependency-injection": "^5.1",
     "symfony/dotenv": "^5.1",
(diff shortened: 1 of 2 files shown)


grasmash:
Weird, why would it work with the patch but not in the new version where the same exact changes are committed?

danepowell:
I think there were other Symfony updates besides just our PRs. One of them appears to have broken Progress Bars.

grasmash:
https://github.com/symfony/symfony/issues/40633

grasmash:
A fix was merged upstream. Waiting for new console release.

anavarre:
Fix is in https://github.com/symfony/symfony/releases/tag/v5.3.0-BETA1 - Do we want to try it out of wait for the final release?

danepowell:
Considering that what blocked us on this PR was a breaking change in a stable patch release, I think the last think we should attempt is update to a beta 😄

danepowell:
There's no need to keep a PR open for this, we'll wait for a new release and track via the internal ticket.

PR #504, opened 2021-04-01
DX-3643: [pull:db] improve memory usage

### LINKED ISSUE #493: Importing a DB into Cloud IDE can be memory intensive
In `/src/Command/Pull/PullCommandBase.php` we have a command to download and import a database.

```
"pv $local_dump_filepath --bytes --rate | gunzip | MYSQL_PWD=$db_password mysql --host=$db_host --user=$db_user $db_name";
```

This operation can have Cloud IDE running out of memory.

```
Choose a Cloud Platform environment:

[0] Dev, dev (vcs: feature-translation-updates)
[1] Dev2, dev2 (vcs: feature-industrial-redesign)
[2] Prod, prod (vcs: tags/2021-03-16.0)
[3] Stage, test (vcs: master)

> 0

[debug] Using environment Dev 29187-bk1618b0-abc8-4115-b62w-775b215ca8ad

[debug] Using database backup generated at 2021-03-17T10:01:18+00:00

[debug] Downloading database backup to /tmp/backup-2021-03-17T10:01:18+00:00-myapp.sql.gz

Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 545015592 bytes) in phar:///opt/local/acli/vendor/guzzlehttp/psr7/src/Stream.php on line 95
```

Not sure how to make this less memory intensive. It's possible that splitting this into 2 commands would be better. E.g., first unzip, then import in a separate command.

danepowell:
**Motivation**

Fixes #493

**Proposed changes**

Instead of reading the contents of the backup file into memory before dumping to disk, just write it to disk as a stream.

**Alternatives considered**

None, this is absolutely the right way to do it 😄 

**Testing steps**

@anavarre can you please help test? I don't have access to any database large enough to reproduce the issue or confirm that it fixes it.

**Merge requirements**

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -281,7 +281,8 @@ abstract class PullCommandBase extends CommandBase {
     if ($response->getStatusCode() !== 200) {
       throw new AcquiaCliException("Unable to download database copy from {$url}. {$response->getStatusCode()}: {$response->getReasonPhrase()}");
     }
-    $backup_file = $response->getBody()->getContents();
+    // Get the response body as a stream to avoid ludicrous memory usage.
+    $backup_file = $response->getBody();
     $this->localMachineHelper->writeFile($local_filepath, $backup_file);
 
     return $local_filepath;
--- a/tests/phpunit/src/CommandTestBase.php
+++ b/tests/phpunit/src/CommandTestBase.php
@@ -377,7 +377,7 @@ abstract class CommandTestBase extends TestBase {
     $backup_id
   ) {
     $stream = $this->prophet->prophesize(StreamInterface::class);
-    $stream->getContents()->willReturn('backupfilecontents');
+    $stream->__toString()->willReturn('backupfilecontents');
     $response = $this->prophet->prophesize(ResponseInterface::class);
     $response->getBody()->willReturn($stream->reveal());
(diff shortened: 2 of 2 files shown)


src/Command/Pull/PullCommandBase.php
```diff
@@ -281,7 +281,8 @@ protected function downloadDatabaseDump(
     if ($response->getStatusCode() !== 200) {
       throw new AcquiaCliException("Unable to download database copy from {$url}. {$response->getStatusCode()}: {$response->getReasonPhrase()}");
     }
-    $backup_file = $response->getBody()->getContents();
+    // Get the response body as a stream to avoid ludicrous memory usage.
+    $backup_file = $response->getBody();
```
grasmash, line 285: I didn’t realize you could just dump the stream to disk.

danepowell: Yeah I had to check the docs, but sure enough file_put_contents (which the Symfony filesystem uses) [accepts a stream resource](https://www.php.net/manual/en/function.file-put-contents.php), and the Guzzle body also [provides a stream resource](https://docs.guzzlephp.org/en/stable/psr7.html#body), so it all just works.

grasmash (changes requested):
Fix tests please :)

PR #506, opened 2021-04-02
[push:code] build artifacts

danepowell:
**Motivation**
Currently, ACLI supports pulling code but not pushing it. Additionally, it has no awareness of the distinction between "source" and "build" repositories. Best practice for Drupal 8+ using Composer is to keep a "source" repo (without dependencies committed, i.e. Github) and an "artifact" repo (with dependencies committed, i.e. Acquia Git). The only way to support this workflow today is with `blt deploy`, which is overly complicated for most users.

**Proposed changes**
Replicate the `blt deploy` command in ACLI, but in a much more streamlined form that's more tightly coupled with Acquia environments, so it's a true deployment manager (not just an artifact builder): allow users to select a specific environment to push to (as long as that environment has a branch checked out).

**Alternatives considered**
Creating a standalone CLI tool to manage deployments so that Acquia CLI remains decoupled from the guts of a Drupal codebase. But this would involve duplicating so much of Acquia CLI, I think it belongs here.

**Testing steps**
3. Clone a Drupal project where vendor dependencies have _not_ been committed. Or create a new project using `composer create-project acquia/drupal-recommended-project`, which should ignore vendor dependencies.
4. Run `acli push:code` and verify that usable code (with production dependencies committed) is deployed to your Acquia environment.

**Merge requirements**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -89,10 +89,14 @@
     "patches": {
       "symfony/console": [
         "https://github.com/symfony/symfony/commit/57910d696b1a3b8573cd562d438880854ba4328e.patch"
+      ],
+      "symfony/filesystem": [
+        "symfony-fs-mirror.patch"
       ]
     },
     "patchLevel": {
-      "symfony/console": "-p5"
+      "symfony/console": "-p5",
+      "symfony/filesystem": "-p5"
     }
   },
   "prefer-stable": true,
--- a/src/Command/Push/PushArtifactCommand.php
+++ b/src/Command/Push/PushArtifactCommand.php
@@ -1,0 +1,242 @@
+<?php
+
+namespace Acquia\Cli\Command\Push;
+
+use Acquia\Cli\Command\Pull\PullCommandBase;
+use Acquia\Cli\Exception\AcquiaCliException;
+use Acquia\Cli\Output\Checklist;
+use Closure;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Finder\Finder;
+use Webmozart\PathUtil\Path;
+
+/**
+ * Class PushArtifactCommand.
+ */
+class PushArtifactCommand extends PullCommandBase {
+
+  protected static $defaultName = 'push:artifact';
+
+  /**
+   * @var string
+   */
+  protected $dir;
+
+  /**
+   * {inheritdoc}.
+   */
+  protected function configure(): void {
+    $this->setDescription('Build and push a code artifact to a Cloud Platform environment')
+      ->addOption('dir', NULL, InputArgument::OPTIONAL, 'The directory containing the Drupal project to be pushed')
+      ->addOption('no-sanitize', NULL, InputOption::VALUE_NONE, 'Do not sanitize the build artifact')
+      ->addOption('dry-run', NULL, InputOption::VALUE_NONE, 'Do not push changes to Acquia Cloud')
+      ->acceptEnvironmentId()
+    ->setHelp("This command builds a sanitized deploy artifact by running <options=bold>composer install</>, removing sensitive files, and committing vendor directories.\n\n"
+      . "The following vendor files and directories are committed to the build artifact even if they are ignored in the source repository: " . implode(', ', self::vendorFiles()) . "\n\n"
+      . "To run additional build or sanitization steps (e.g. <options=bold>npm install</>), add a <options=bold>post-install-cmd</> script to your <options=bold>composer.json</> file: https://getcomposer.org/doc/articles/scripts.md#command-events");
+  }
+
+  /**
+   * @param \Symfony\Component\Console\Input\InputInterface $input
+   * @param \Symfony\Component\Console\Output\OutputInterface $output
+   *
+   * @return int 0 if everything went fine, or an exit code
+   * @throws \Exception
+   */
+  protected function execute(InputInterface $input, OutputInterface $output): int {
+    // @todo change deploy strategy depending on whether this is a "source" repo (requiring building) or not
+    // @todo handle if Git user name/email is missing
+    $this->setDirAndRequireProjectCwd($input);
+    $is_dirty = $this->isLocalGitRepoDirty();
+    $commit_hash = $this->getLocalGitCommitHash();
+    if ($is_dirty) {
+      throw new AcquiaCliException('Pushing code was aborted because your local Git repository has uncommitted changes. Please either commit, reset, or stash your changes via git.');
+    }
+    $this->checklist = new Checklist($output);
+
+    // @todo handle environments with tags deployed
+    $this->logger->warning('You must select an environment with a Git branch deployed');
+    $environment = $this->determineEnvironment($input, $output, TRUE);
+    if (strpos($environment->vcs->path, 'tags') === 0) {
+      throw new AcquiaCliException("Environments with Git tags are not supported. Environment {$environment->name} has {$environment->vcs->path} deployed.");
+    }
+    $artifact_dir = Path::join(sys_get_temp_dir(), 'acli-push-artifact');
+    $output_callback = $this->getOutputCallback($output, $this->checklist);
+
+    $this->checklist->addItem('Preparing artifact directory');
+    $this->prepareDir($output_callback, $artifact_dir, $environment->vcs->url, $environment->vcs->path);
+    $this->checklist->completePreviousItem();
+
+    $this->checklist->addItem('Generating build artifact');
+    $this->build($output_callback, $artifact_dir);
+    $this->checklist->completePreviousItem();
+
+    if (!$input->getOption('no-sanitize')) {
+      $this->checklist->addItem('Sanitizing build artifact');
+      $this->sanitize($artifact_dir);
+      $this->checklist->completePreviousItem();
+    }
+
+    $this->checklist->addItem("Committing changes (commit hash: $commit_hash)");
+    $this->commit($output_callback, $artifact_dir, $commit_hash);
+    $this->checklist->completePreviousItem();
+
+    if (!$input->getOption('dry-run')) {
+      $this->checklist->addItem("Pushing changes to {$environment->vcs->path} branch in the {$environment->name} environment");
+      $this->push($output_callback, $artifact_dir);
+      $this->checklist->completePreviousItem();
+    }
+    else {
+      $this->logger->warning("The <options=bold>--dry-run</> option prevented changes from being pushed to Acquia Cloud. The artifact has been built at $artifact_dir");
+    }
+
+    return 0;
+  }
+
+  /**
+   * Prepares a directory to build the artifact.
+   *
+   * @param \Closure $output_callback
+   * @param string $artifact_dir
+   * @param string $vcs_url
+   * @param string $vcs_path
+   *
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function prepareDir(Closure $output_callback, string $artifact_dir, string $vcs_url, string $vcs_path): void {
+    $fs = $this->localMachineHelper->getFilesystem();
+
+    $this->logger->info("Removing $artifact_dir if it exists");
+    $fs->remove($artifact_dir);
+
+    $this->logger->info("Initializing Git in $artifact_dir");
+    $process = $this->localMachineHelper->executeFromCmd("git clone --depth 1 --branch $vcs_path $vcs_url $artifact_dir", $output_callback, NULL, $this->output->isVerbose());
+    if (!$process->isSuccessful()) {
+      throw new AcquiaCliException('Failed to clone repository from the Cloud Platform: {message}', ['message' => $process->getErrorOutput()]);
+    }
+
+    $this->logger->info('Global .gitignore file is temporarily disabled during artifact builds.');
+    $this->localMachineHelper->executeFromCmd('git config --local core.excludesFile false', $output_callback, $artifact_dir, $this->output->isVerbose());
+    $this->localMachineHelper->executeFromCmd('git config --local core.fileMode true', $output_callback, $artifact_dir, $this->output->isVerbose());
+
+    // Vendor directories can be "corrupt" (i.e. missing scaffold files due to earlier sanitization) in ways that break composer install.
+    $this->logger->info('Removing vendor directories');
+    foreach (self::vendorFiles() as $vendor_directory) {
+      $fs->remove(Path::join($artifact_dir, $vendor_directory));
+    }
+  }
+
+  protected function build(Closure $output_callback, string $artifact_dir): void {
+    // @todo generate a deploy identifier
+    // @see https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295
+    $this->logger->info("Mirroring source files from {$this->dir} to $artifact_dir");
+    $originFinder = Finder::create();
+    $originFinder->files()->in($this->dir)
+      // Include dot files like .htaccess.
+      ->ignoreDotFiles(FALSE)
+      // Ignore VCS ignored files (e.g. vendor) to speed up the mirror (Composer will restore them later).
+      ->ignoreVCSIgnored(TRUE);
+    $targetFinder = Finder::create();
+    $targetFinder->files()->in($artifact_dir)->ignoreDotFiles(FALSE);
+    $this->localMachineHelper->getFilesystem()->mirror($this->dir, $artifact_dir, $originFinder, ['override' => TRUE, 'delete' => TRUE], $targetFinder);
+
+    $this->logger->info('Installing Composer production dependencies');
+    $this->localMachineHelper->executeFromCmd('composer install --no-dev --no-interaction --optimize-autoloader', $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function sanitize(string $artifact_dir):void {
+    $this->logger->info('Finding Drupal core text files');
+    $sanitizeFinder = Finder::create()
+      ->files()
+      ->name('*.txt')
+      ->notName('LICENSE.txt')
+      ->in("$artifact_dir/docroot/core");
+
+    $this->logger->info('Finding VCS directories');
+    $vcsFinder = Finder::create()
+      ->ignoreDotFiles(FALSE)
+      ->ignoreVCS(FALSE)
+      ->directories()
+      ->in(["$artifact_dir/docroot",
+        "$artifact_dir/vendor",
+      ])
+      ->name('.git');
+    $drush_dir = "$artifact_dir/drush";
+    if (file_exists($drush_dir)) {
+      $vcsFinder->in($drush_dir);
+    }
+    if ($vcsFinder->hasResults()) {
+      $sanitizeFinder->append($vcsFinder);
+    }
+
+    $this->logger->info('Finding INSTALL database text files');
+    $dbInstallFinder = Finder::create()
+      ->files()
+      ->in([$artifact_dir])
+      ->name('/INSTALL\.[a-z]+\.(md|txt)$/');
+    if ($dbInstallFinder->hasResults()) {
+      $sanitizeFinder->append($dbInstallFinder);
+    }
+
+    $this->logger->info('Finding other common text files');
+    $filenames = [
+      'AUTHORS',
+      'CHANGELOG',
+      'CONDUCT',
+      'CONTRIBUTING',
+      'INSTALL',
+      'MAINTAINERS',
+      'PATCHES',
+      'TESTING',
+      'UPDATE',
+    ];
+    $textFileFinder = Finder::create()
+      ->files()
+      ->in(["$artifact_dir/docroot"])
+      ->name('/(' . implode('|', $filenames) . ')\.(md|txt)$/');
+    if ($textFileFinder->hasResults()) {
+      $sanitizeFinder->append($textFileFinder);
+    }
+
+    $this->logger->info("Removing sanitized files from build");
+    $this->localMachineHelper->getFilesystem()->remove($sanitizeFinder);
+  }
+
+  protected function commit(Closure $output_callback, string $artifact_dir, string $commit_hash):void {
+    $this->logger->info('Adding and committing changed files');
+    $this->localMachineHelper->executeFromCmd('git add -A', $output_callback, $artifact_dir, $this->output->isVerbose());
+    foreach (self::vendorFiles() as $vendor_directory) {
+      // This will fatally error if the directory doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $vendor_directory");
+      $this->localMachineHelper->executeFromCmd('git add -f ' . $vendor_directory, NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->executeFromCmd('git commit -m "Automated commit by Acquia CLI (source commit: ' . $commit_hash . ')"', $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function push(Closure $output_callback, string $artifact_dir):void {
+    $this->logger->info('Pushing changes to Acquia Git');
+    $this->localMachineHelper->executeFromCmd('git push', $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  private static function vendorFiles(): array {
+    return [
+      'vendor',
+      'docroot/core',
+      'docroot/modules/contrib',
+      'docroot/themes/contrib',
+      'docroot/profiles/contrib',
+      'docroot/libraries',
+      'drush/Commands',
+      'docroot/index.php',
+      'docroot/.htaccess',
+      'docroot/autoload.php',
+      'docroot/robots.txt',
+      'docroot/update.php',
+      'docroot/web.config'
+    ];
+  }
+
+}
--- a/src/Command/Push/PushArtifactCommand.php
+++ b/src/Command/Push/PushArtifactCommand.php
@@ -76,7 +76,7 @@ class PushArtifactCommand extends PullCommandBase {
 
     if (!$input->getOption('no-sanitize')) {
       $this->checklist->addItem('Sanitizing build artifact');
-      $this->sanitize($artifact_dir);
+      $this->sanitize($output_callback, $artifact_dir);
       $this->checklist->completePreviousItem();
     }
 
@@ -109,21 +109,21 @@ class PushArtifactCommand extends PullCommandBase {
   protected function prepareDir(Closure $output_callback, string $artifact_dir, string $vcs_url, string $vcs_path): void {
     $fs = $this->localMachineHelper->getFilesystem();
 
-    $this->logger->info("Removing $artifact_dir if it exists");
+    $output_callback('out', "Removing $artifact_dir if it exists");
     $fs->remove($artifact_dir);
 
-    $this->logger->info("Initializing Git in $artifact_dir");
+    $output_callback('out', "Initializing Git in $artifact_dir");
     $process = $this->localMachineHelper->executeFromCmd("git clone --depth 1 --branch $vcs_path $vcs_url $artifact_dir", $output_callback, NULL, $this->output->isVerbose());
     if (!$process->isSuccessful()) {
       throw new AcquiaCliException('Failed to clone repository from the Cloud Platform: {message}', ['message' => $process->getErrorOutput()]);
     }
 
-    $this->logger->info('Global .gitignore file is temporarily disabled during artifact builds.');
+    $output_callback('out', 'Global .gitignore file is temporarily disabled during artifact builds.');
     $this->localMachineHelper->executeFromCmd('git config --local core.excludesFile false', $output_callback, $artifact_dir, $this->output->isVerbose());
     $this->localMachineHelper->executeFromCmd('git config --local core.fileMode true', $output_callback, $artifact_dir, $this->output->isVerbose());
 
     // Vendor directories can be "corrupt" (i.e. missing scaffold files due to earlier sanitization) in ways that break composer install.
-    $this->logger->info('Removing vendor directories');
+    $output_callback('out', 'Removing vendor directories');
     foreach (self::vendorFiles() as $vendor_directory) {
       $fs->remove(Path::join($artifact_dir, $vendor_directory));
     }
@@ -132,7 +132,7 @@ class PushArtifactCommand extends PullCommandBase {
   protected function build(Closure $output_callback, string $artifact_dir): void {
     // @todo generate a deploy identifier
     // @see https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295
-    $this->logger->info("Mirroring source files from {$this->dir} to $artifact_dir");
+    $output_callback('out', "Mirroring source files from {$this->dir} to $artifact_dir");
     $originFinder = Finder::create();
     $originFinder->files()->in($this->dir)
       // Include dot files like .htaccess.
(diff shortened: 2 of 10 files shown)


danepowell:
@grasmash @anavarre would you mind taking this for a spin and providing preliminary feedback on the functionality and code? If you're good with the general approach, I'll clean it up and add tests.

composer.json
```diff
@@ -89,10 +89,14 @@
     "patches": {
       "symfony/console": [
         "https://github.com/symfony/symfony/commit/57910d696b1a3b8573cd562d438880854ba4328e.patch"
+      ],
+      "symfony/filesystem": [
+        "symfony-fs-mirror.patch"
```
danepowell, line 94: https://github.com/symfony/symfony/pull/40689

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +40,157 @@ protected function configure() {
...
+    $this->logger->info('Global .gitignore file is temporarily disabled during artifact builds.');
+    $this->localMachineHelper->executeFromCmd('git config --local core.excludesFile false', $output_callback, $artifact_dir, $this->output->isVerbose());
+    $this->localMachineHelper->executeFromCmd('git config --local core.fileMode true', $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function build(Closure $output_callback, string $artifact_dir): void {
+    // @todo generate a deploy identifier
+    // @see https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295
+
+    // @todo support user-defined scripts such as npm build, or recommend how to run them via Composer
```
grasmash, line 109: I would strongly prefer to leverage Composer here.

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +40,157 @@ protected function configure() {
...
+    ];
+    $textFileFinder = Finder::create()
+      ->files()
+      ->in(["$artifact_dir/docroot"])
+      ->name('/(' . implode('|', $filenames) . ')\.(md|txt)$/');
+    if ($textFileFinder->hasResults()) {
+      $sanitizeFinder->append($textFileFinder);
+    }
+
+    // @todo support custom sanitization as with BLT?
```
grasmash, line 182: At most, I would support running a Composer script like “composer sanitize”

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +40,157 @@ protected function configure() {
...
+
+    $this->logger->info("Mirroring source files from {$this->dir} to $artifact_dir");
+    $originFinder = Finder::create();
+    $originFinder->files()->in($this->dir)
+      // Include dot files like .htaccess.
+      ->ignoreDotFiles(FALSE)
+      // Ignore VCS ignored files (e.g. vendor) to speed up the mirror (Composer will restore them later).
+      ->ignoreVCSIgnored(TRUE)
+      // Ignore .gitignore files (everything should be committed to the artifact).
+      ->notName('.gitignore');
```
grasmash, line 119: I wonder if we should just force commit a handful of directory us rather than everything.

grasmash:
When Acquia Cloud Platform eventually supports tasks that perform Composer assembly, I'd like that task to simply run an Acquia CLI command.

For that reason, it's likely best to allow a user to generate a deployment artifact without pushing it. E.g., run `acli push:code --dry-run` or `acli push:code --build-only`.

src/Command/Push/PushCodeCommand.php
grasmash: E.g., run `composer run-script sanitize`

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +40,157 @@ protected function configure() {
...
+    $targetFinder = Finder::create();
+    $targetFinder->files()->in($artifact_dir)->ignoreDotFiles(FALSE);
+    $this->localMachineHelper->getFilesystem()->mirror($this->dir, $artifact_dir, $originFinder, ['override' => TRUE, 'delete' => TRUE], $targetFinder);
+
+    $this->logger->info('Installing Composer production dependencies');
+    $this->localMachineHelper->executeFromCmd('composer install --no-dev --no-interaction --optimize-autoloader', $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function sanitize(string $artifact_dir):void {
```
grasmash, line 128: We may also want to make sanitization optional by adding a `--no-sanitize` flag.

src/Command/Push/PushCodeCommand.php
danepowell: You're right, we can't be so naïve as to commit _everything_, because that'll include `sites/default/files`. Hmm...

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +41,160 @@ protected function configure() {
...
+
+  protected function push(Closure $output_callback, string $artifact_dir, string $commit_hash):void {
+    $this->logger->info('Adding and committing changed files');
+    $force_add = [
+      'vendor',
+      'docroot/core',
+      'docroot/modules/contrib',
+      'docroot/themes/contrib',
+      'docroot/profiles/contrib',
+      'docroot/libraries'
```
grasmash, line 190: Drush commands too.

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +41,160 @@ protected function configure() {
...
+    ];
+    $this->localMachineHelper->executeFromCmd('git add -A', $output_callback, $artifact_dir, $this->output->isVerbose());
+    $this->localMachineHelper->executeFromCmd('git add -f ' . implode(' ', $force_add), $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 193: I think it'd be good to emit a message showing exactly what we're force committing.

danepowell:
Let me know if you have any other feedback, or if it's time to write tests

grasmash:
I'm concerned that this command does a lot more than the name implies. `acli push:code` does not imply a "build and push" operation. I think we need to make that very clear so that the user knows what process they are initiating.

src/Command/Push/PushCodeCommand.php
```diff
@@ -34,10 +46,191 @@ protected function configure() {
...
+
+  protected function commit(Closure $output_callback, string $artifact_dir, string $commit_hash):void {
+    $this->logger->info('Adding and committing changed files');
+    $this->localMachineHelper->executeFromCmd('git add -A', $output_callback, $artifact_dir, $this->output->isVerbose());
+    foreach (self::vendorDirectories() as $vendor_directory) {
+      // This will fatally error if the directory doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $vendor_directory");
+      $this->localMachineHelper->executeFromCmd('git add -f ' . $vendor_directory, NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->executeFromCmd('git commit -m "Automated commit by Acquia CLI (source commit: ' . $commit_hash . ')"', $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 216: Just thinking this through, we're committing to the current branch? That seems a bit dangerous, right? We're effectively transforming a "source branch" into a "build branch."

danepowell:
Apparently many other files such as `docroot/index.php` can be gitignored. I don't know whether to add these to the list to force-commit (which is going to grow to several dozen files at least), or take some other approach.

src/Command/Push/PushCodeCommand.php
danepowell: We're committing to the branch currently checked out on Acquia Cloud. So it's definitely going to be a build branch. All of these commands are running in the artifact directory, which is checked out from cloud.

ACLI remains completely unaware of your "source" branch, it just mirrors whatever code is in the directory from which you run it.

danepowell:
> I'm concerned that this command does a lot more than the name implies

It does, but I think the alternative (_just_ pushing code without any sort of intelligent build process or deployment) is likely to lead to more confusion. This is what killed BLT over and over: when people want to deploy, they mostly want to deploy _to_ some environment. Making deployment a two- or three-step process (building, committing to Git, deploying Git to a specific environment), while technically somewhat cleaner, generated endless frustration, as measured by support tickets.

And given that there is no other canonical way of building a Drupal deploy artifact, I see no alternative but to handle this in ACLI.

grasmash:
> I'm concerned that this command does a lot more than the name implies. `acli push:code` does not imply a "build and push" operation. I think we need to make that very clear so that the user knows what process they are initiating.

Would you consider renaming this to something that's more obvious, like `acli push:artifact` or `acli:push-build` ?

danepowell:
> Would you consider renaming this to something that's more obvious

That's fine, although eventually I think we shouldn't force users to consider the distinction between source and build artifacts. They should just "push" and get a working deployment. Even as-is, this command should work equally well on a source or build artifact, people might just be surprised by the sanitization.

danepowell:
@grasmash @anavarre am I safe to write tests for this? i.e. are we confident in the overall approach?

anavarre:
I think that's a great idea and very useful for our customers. Thanks for doing it!

Confirmed that:

- Expanding the Travis log file, I can file the transfer.sh unique link
- Clicking on the link brings me to a transfer.sh download page and the download works fine
- Checking ACLI's version locally then gave me `1.9.0-14-g78979c9` which is exactly at the SHA1 for the Travis build

anavarre:
> am I safe to write tests for this? i.e. are we confident in the overall approach?

I don't think I'm qualified to say yay or nay here. Based on your own understanding of `blt deploy` I think that should be a decision between you and Matt. I certainly appreciate the `push:code` overhaul and think customers will find value in this, but I don't yet envision whether it's something that customers will use a lot, so will rely on your own understanding of how `blt deploy` is currently used (and its limitations).

grasmash:
Yes, I think so.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * {inheritdoc}.
+   */
+  protected function configure(): void {
+    $this->setDescription('Build and push a code artifact to a Cloud Platform environment')
+      ->addOption('dir', NULL, InputArgument::OPTIONAL, 'The directory containing the Drupal project to be pushed')
+      ->addOption('no-sanitize', NULL, InputOption::VALUE_NONE, 'Do not sanitize the build artifact')
+      ->addOption('dry-run', NULL, InputOption::VALUE_NONE, 'Do not push changes to Acquia Cloud')
+      ->acceptEnvironmentId()
+    ->setHelp("This command builds a sanitized deploy artifact by running <options=bold>composer install</>, removing sensitive files, and committing vendor directories.\n\n"
```
grasmash, line 37: I'd prefer using `PHP_EOL` rather than "\n"

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+    $this->setDirAndRequireProjectCwd($input);
+    $is_dirty = $this->isLocalGitRepoDirty();
+    $commit_hash = $this->getLocalGitCommitHash();
+    if ($is_dirty) {
+      throw new AcquiaCliException('Pushing code was aborted because your local Git repository has uncommitted changes. Please either commit, reset, or stash your changes via git.');
+    }
+    $this->checklist = new Checklist($output);
+
+    // @todo handle environments with tags deployed
+    $this->logger->warning('You must select an environment with a Git branch deployed');
```
grasmash, line 61: Why are we showing a warning here? Seems like this is more informational, part of the context for the prompt.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+    $environment = $this->determineEnvironment($input, $output, TRUE);
+    if (strpos($environment->vcs->path, 'tags') === 0) {
+      throw new AcquiaCliException("Environments with Git tags are not supported. Environment {$environment->name} has {$environment->vcs->path} deployed.");
```
grasmash, line 64:
```suggestion
      throw new AcquiaCliException("You cannot push to an environment that has a git tag deployed to it. Environment {$environment->name} has {$environment->vcs->path} deployed. Please select a different environment.");
```

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+    $this->commit($output_callback, $artifact_dir, $commit_hash);
+    $this->checklist->completePreviousItem();
+
+    if (!$input->getOption('dry-run')) {
+      $this->checklist->addItem("Pushing changes to {$environment->vcs->path} branch in the {$environment->name} environment");
+      $this->push($output_callback, $artifact_dir);
+      $this->checklist->completePreviousItem();
+    }
+    else {
+      $this->logger->warning("The <options=bold>--dry-run</> option prevented changes from being pushed to Acquia Cloud. The artifact has been built at $artifact_dir");
```
grasmash, line 93:
```suggestion
      $this->logger->warning("The <options=bold>--dry-run</> option prevented changes from being pushed to Acquia Cloud. The artifact has been built at <options=bold>$artifact_dir</>");
```

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+      $this->sanitize($artifact_dir);
+      $this->checklist->completePreviousItem();
+    }
+
+    $this->checklist->addItem("Committing changes (commit hash: $commit_hash)");
+    $this->commit($output_callback, $artifact_dir, $commit_hash);
+    $this->checklist->completePreviousItem();
+
+    if (!$input->getOption('dry-run')) {
+      $this->checklist->addItem("Pushing changes to {$environment->vcs->path} branch in the {$environment->name} environment");
```
grasmash, line 88:
```suggestion
      $this->checklist->addItem("Pushing changes to <options=bold>{$environment->vcs->path}</> branch in the <options=bold>{$environment->name}</> environment");
```

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+   * @param string $artifact_dir
+   * @param string $vcs_url
+   * @param string $vcs_path
+   *
+   * @throws \Acquia\Cli\Exception\AcquiaCliException
+   */
+  protected function prepareDir(Closure $output_callback, string $artifact_dir, string $vcs_url, string $vcs_path): void {
+    $fs = $this->localMachineHelper->getFilesystem();
+
+    $this->logger->info("Removing $artifact_dir if it exists");
```
grasmash, line 112: Use `$this->io`

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+    $fs->remove($artifact_dir);
+
+    $this->logger->info("Initializing Git in $artifact_dir");
+    $process = $this->localMachineHelper->executeFromCmd("git clone --depth 1 --branch $vcs_path $vcs_url $artifact_dir", $output_callback, NULL, $this->output->isVerbose());
```
grasmash, line 116: Seems better to call `execute($command)` rather than passing a string to `executeFromCmd()`

src/Command/Push/PushArtifactCommand.php
grasmash: Same with following lines.

grasmash: Actually, it seems like this should be added to the `$output_callback` via `$output_callback('out', "Removing $artifact_dir if it exists")`

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  private static function vendorFiles(): array {
+    return [
+      'vendor',
+      'docroot/core',
+      'docroot/modules/contrib',
+      'docroot/themes/contrib',
+      'docroot/profiles/contrib',
+      'docroot/libraries',
```
grasmash, line 231: Seems like we may want to read some of these values from composer.json.

grasmash (changes requested):

src/Command/Push/PushArtifactCommand.php
danepowell: Indeed it's just informational, but I could not find any less severe log method that would display in non-verbose output. Am I missing something?

danepowell: Oh duh, I should be using `$output`, not `$this->logger`

src/Command/Push/PushArtifactCommand.php
danepowell: Great idea!

danepowell:
Working on tests now, it'll take a while. I wonder if I can use `LocalMachineHelper::readFile()` instead of `file_get_contents()` and then mock `readFile()` to avoid having to create a whole fixture including Drupal core.

danepowell:
Alright, this is good to go if y'all approve.

> Based on your own understanding of blt deploy

This is a pretty major improvement over `blt deploy` in several ways. It's a lot simpler at every step, and smarter when it comes to including vendor/scaffold files in the artifact, no longer requiring giant hardcoded lists of files to add, exclude, or ignore. It sacrifices a little flexibility in the sanitization step but we can deal with that later if it's really in demand.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    $this->setDirAndRequireProjectCwd($input);
+    $is_dirty = $this->isLocalGitRepoDirty();
+    $commit_hash = $this->getLocalGitCommitHash();
+    if ($is_dirty) {
+      throw new AcquiaCliException('Pushing code was aborted because your local Git repository has uncommitted changes. Please either commit, reset, or stash your changes via git.');
+    }
+    $this->checklist = new Checklist($output);
+
+    // @todo handle environments with tags deployed
+    $output->writeln('<info>You must select an environment with a Git branch deployed</info>');
```
grasmash, line 77: We should use IO

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    $this->commit($output_callback, $artifact_dir, $commit_hash);
+    $this->checklist->completePreviousItem();
+
+    if (!$input->getOption('dry-run')) {
+      $this->checklist->addItem("Pushing changes to <options=bold>{$environment->vcs->path}</> branch in the <options=bold>{$environment->name}</> environment");
+      $this->push($output_callback, $artifact_dir);
+      $this->checklist->completePreviousItem();
+    }
+    else {
+      $this->logger->warning("The <options=bold>--dry-run</> option prevented changes from being pushed to Acquia Cloud. The artifact has been built at <options=bold>$artifact_dir</>");
```
grasmash, line 109: Use IO

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    $this->localMachineHelper->execute(['git', 'add', '-A'], $output_callback, $artifact_dir, $this->output->isVerbose());
+    foreach (array_merge($this->vendorDirs($artifact_dir), $this->scaffoldFiles($artifact_dir)) as $file) {
+      // This will fatally error if the file doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $file");
+      $this->localMachineHelper->execute(['git', 'add', '-f', $file], NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->execute(['git', 'commit', '-m', "Automated commit by Acquia CLI (source commit: $commit_hash)"], $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function push(Closure $output_callback, string $artifact_dir):void {
```
grasmash, line 235: I’m surprised phpcs passes without method comments.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+   *
+   * Composer scaffold files.
+   */
+  protected $scaffoldFiles;
+
+  /**
+   * {inheritdoc}.
+   */
+  protected function configure(): void {
+    $this->setDescription('Build and push a code artifact to a Cloud Platform environment')
```
grasmash, line 48: It would be helpful to add a couple of example usages

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+
+  protected function commit(Closure $output_callback, string $artifact_dir, string $commit_hash):void {
+    $output_callback('out', 'Adding and committing changed files');
+    $this->localMachineHelper->execute(['git', 'add', '-A'], $output_callback, $artifact_dir, $this->output->isVerbose());
+    foreach (array_merge($this->vendorDirs($artifact_dir), $this->scaffoldFiles($artifact_dir)) as $file) {
+      // This will fatally error if the file doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $file");
+      $this->localMachineHelper->execute(['git', 'add', '-f', $file], NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->execute(['git', 'commit', '-m', "Automated commit by Acquia CLI (source commit: $commit_hash)"], $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 232: It seems like we are adding a new System dependency for git. should we check for it before we start the process?

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+      $sanitizeFinder->append($textFileFinder);
+    }
+
+    $output_callback('out', "Removing sanitized files from build");
+    $this->localMachineHelper->getFilesystem()->remove($sanitizeFinder);
+  }
+
+  protected function commit(Closure $output_callback, string $artifact_dir, string $commit_hash):void {
+    $output_callback('out', 'Adding and committing changed files');
+    $this->localMachineHelper->execute(['git', 'add', '-A'], $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 226: Will this add new files or only the currently tracked ones?

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    ];
+    $textFileFinder = $this->localMachineHelper->getFinder()
+      ->files()
+      ->in(["$artifact_dir/docroot"])
+      ->name('/(' . implode('|', $filenames) . ')\.(md|txt)$/');
+    if ($textFileFinder->hasResults()) {
+      $sanitizeFinder->append($textFileFinder);
+    }
+
+    $output_callback('out', "Removing sanitized files from build");
```
grasmash, line 220: I don’t think that the files themselves are sanitized.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+      // Include dot files like .htaccess.
+      ->ignoreDotFiles(FALSE)
+      // Ignore VCS ignored files (e.g. vendor) to speed up the mirror (Composer will restore them later).
+      ->ignoreVCSIgnored(TRUE);
+    $targetFinder = $this->localMachineHelper->getFinder();
+    $targetFinder->files()->in($artifact_dir)->ignoreDotFiles(FALSE);
+    $this->localMachineHelper->getFilesystem()->mirror($this->dir, $artifact_dir, $originFinder, ['override' => TRUE, 'delete' => TRUE], $targetFinder);
+
+    $output_callback('out', 'Installing Composer production dependencies');
+    $this->localMachineHelper->execute(['composer', 'install', '--no-dev', '--no-interaction', '--optimize-autoloader'], $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 163: This is another new system dependency

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+  protected function prepareDir(Closure $output_callback, string $artifact_dir, string $vcs_url, string $vcs_path): void {
+    $fs = $this->localMachineHelper->getFilesystem();
+
+    $output_callback('out', "Removing $artifact_dir if it exists");
+    $fs->remove($artifact_dir);
+
+    $output_callback('out', "Initializing Git in $artifact_dir");
+    $process = $this->localMachineHelper->execute(['git', 'clone', '--depth', '1', '--branch', $vcs_path, $vcs_url, $artifact_dir], $output_callback, NULL, $this->output->isVerbose());
+    if (!$process->isSuccessful()) {
+      throw new AcquiaCliException('Failed to clone repository from the Cloud Platform: {message}', ['message' => $process->getErrorOutput()]);
```
grasmash, line 134: You may want to include the branch information here too

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+      // This will fatally error if the file doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $file");
+      $this->localMachineHelper->execute(['git', 'add', '-f', $file], NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->execute(['git', 'commit', '-m', "Automated commit by Acquia CLI (source commit: $commit_hash)"], $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function push(Closure $output_callback, string $artifact_dir):void {
+    $output_callback('out', 'Pushing changes to Acquia Git');
+    $this->localMachineHelper->execute(['git', 'push'], $output_callback, $artifact_dir, $this->output->isVerbose());
```
grasmash, line 237: The repository in the current working directory will be left in a strange state after this command executes. There will be a new upstream commit resulting from this command, but the local repository won’t have it. Perhaps we should pull it down? Otherwise, we could at least inform the user.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    foreach (array_merge($this->vendorDirs($artifact_dir), $this->scaffoldFiles($artifact_dir)) as $file) {
+      // This will fatally error if the file doesn't exist. Suppress error output.
+      $this->logger->debug("Forcibly adding $file");
+      $this->localMachineHelper->execute(['git', 'add', '-f', $file], NULL, $artifact_dir, FALSE);
+    }
+    $this->localMachineHelper->execute(['git', 'commit', '-m', "Automated commit by Acquia CLI (source commit: $commit_hash)"], $output_callback, $artifact_dir, $this->output->isVerbose());
+  }
+
+  protected function push(Closure $output_callback, string $artifact_dir):void {
+    $output_callback('out', 'Pushing changes to Acquia Git');
```
grasmash, line 236: It would be informative to specify the specific remote that is being pushed to

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    $is_dirty = $this->isLocalGitRepoDirty();
+    $commit_hash = $this->getLocalGitCommitHash();
+    if ($is_dirty) {
+      throw new AcquiaCliException('Pushing code was aborted because your local Git repository has uncommitted changes. Please either commit, reset, or stash your changes via git.');
+    }
+    $this->checklist = new Checklist($output);
+
+    // @todo handle environments with tags deployed
+    $output->writeln('<info>You must select an environment with a Git branch deployed</info>');
+    $environment = $this->determineEnvironment($input, $output, TRUE);
```
grasmash, line 78: It seems like this command doesn’t actually push up the local work. Instead, it checks out a remote branch, builds  it into an artifact, and pushes it back. That’s not what I would expect from the command. I would expect there to be local changes which are not upstream. based on those local changes, an artifact would be created and pushed. the end up getting pushed up stream.

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    if (strpos($environment->vcs->path, 'tags') === 0) {
+      throw new AcquiaCliException("You cannot push to an environment that has a git tag deployed to it. Environment {$environment->name} has {$environment->vcs->path} deployed. Please select a different environment.");
```
grasmash, line 80: This is a possibly strange assumption. Why do we care about what is checked out on the environment? Are we pushing to remote or are we pushing to an environment? If we are pushing to a remote and it doesn’t matter. If we are pushing to an environment, why can’t we check out a different branch or tag as part of that push?

src/Command/Push/PushArtifactCommand.php
```diff
@@ -0,0 +1,284 @@
...
+    $output_callback('out', 'Removing vendor directories');
+    foreach ($this->vendorDirs($artifact_dir) as $vendor_directory) {
+      $fs->remove(Path::join($artifact_dir, $vendor_directory));
+    }
+  }
+
+  protected function build(Closure $output_callback, string $artifact_dir): void {
+    // @todo generate a deploy identifier
+    // @see https://git.drupalcode.org/project/drupal/-/blob/9.1.x/sites/default/default.settings.php#L295
+    $output_callback('out', "Mirroring source files from {$this->dir} to $artifact_dir");
```
grasmash, line 151: I think there’s a scenario in which this is problematic. If there is a new upstream change to the repository which has not yet been pulled locally, and then a user runs this command, it will effectively wipe out those changes. What is the desired behavior in that case?

it would be nice to check whether the local repository is up-to-date before initiating this build.

grasmash (changes requested):

src/Command/Push/PushArtifactCommand.php
grasmash: In other words, we should reword this a bit

src/Command/Push/PushArtifactCommand.php
grasmash: In other words, we should reword this a bit

src/Command/Push/PushArtifactCommand.php
grasmash: we probably should check out the pushed branch on the remote environment. I’m fine with preventing that for environments that have tags, but it only makes sense if we are otherwise going to be potentially deploying a new branch.

src/Command/Push/PushArtifactCommand.php
danepowell: I don't see how to use IO here. When I tried, it wasn't interpreting `<options=bold></>`. In other areas of the code where we have a warning and font styles, we use `$this->logger->warning()`.

src/Command/Push/PushArtifactCommand.php
danepowell: I can change it. But... between `$this->logger`, `$this->io`, `$output`, and `$output_callback`, all of which largely overlap, I am completely lost as to what to use when.

Since you seem to have strong opinions on this, would you mind leaving a comment here or in [our style guide](https://github.com/acquia/cli/blob/master/CONTRIBUTING.md#style-guide) so we can get it right the first time next time?

src/Command/Push/PushArtifactCommand.php
danepowell: Oh boy, have we been down this road before 😄 Not sure if you remember the conversation on RA. Coder only complains if comments are _wrong_, it doesn't care about _missing_ comments.

src/Command/Push/PushArtifactCommand.php
danepowell: With only two options it's not that complex, and we already get the standard examples out of the box:
```
Usage:
  push:artifact [options] [--] [<environmentId>]
  push:artifact [<environmentAlias>]
  push:artifact myapp.dev
  push:artifact 12345-abcd1234-1111-2222-3333-0e02b2c3d470
```

But I can add another.

src/Command/Push/PushArtifactCommand.php
danepowell: `LocalMachineHelper::commandExists()` isn't very reliable in its current state. I think checking for a common dependency like Git is more likely to do harm than good. I think we have tickets to both improve `LocalMachineHelper::commandExists()` and to create a Git wrapper, those would be the place to add a check for Git.

src/Command/Push/PushArtifactCommand.php
danepowell: Au contraire, we use it in `pull` and `new`. As with Git, I agree we'd ideally check for it, but we can't do that before fixing `commandExists()` (it's _super_ common for `composer` to be an alias, in fact Acquia recommends it as a workaround for Composer 1/2 issues)

src/Command/Push/PushArtifactCommand.php
danepowell: I don't understand. Are you confusing the source and artifact repos? These commands all execute in a dedicated artifact directory (`/tmp/acli-build-artifact`) that the user will never see. We never modify the source repo (i.e. the user's working directory).

src/Command/Push/PushArtifactCommand.php
danepowell: I want the user to think of deploys as being to an _environment_, not to a particular branch or tag. This was biggest point of confusion with `blt deploy`, people didn't understand that they had to deploy to _git_ first, and then deploy git to an environment.

This is easy to accomplish technically if an environment is running a branch, we just push to the branch. It's less clear what we should do if an environment runs a tag. I'm kicking that decision down the road in the spirit of getting this out as an MVP.

danepowell:
Per discussion in Slack, this is okay to merge with the follow-up tickets we've created:
- https://github.com/acquia/cli/issues/530
- https://github.com/acquia/cli/issues/525

I wish we could mark a command as "beta" or something. I think it's really important to get this out to start getting feedback, but I agree that it's a little more complex and likely to cause confusion.

PR #516, opened 2021-04-09
Fetch hosting alias correctly.

grasmash:


src/Command/CommandBase.php
```diff
@@ -1089,7 +1089,7 @@ public static function getAliasCache() {
    */
   protected function doGetApplicationFromAlias($application_alias) {
     $acquia_cloud_client = $this->cloudApiClientService->getClient();
-    $acquia_cloud_client->addQuery('filter', 'hosting=@*' . $application_alias);
+    $acquia_cloud_client->addQuery('filter', 'hosting=prod:' . $application_alias);
```
grasmash, line 1092: Sometimes this wildcard causes multiple applications to be returned. E.g., when an alias is `blue` and there are apps `prod:greenblue, prod:veryblue, prod:blue`, then `prod:greenblue` may be returned when `blue` is requested.

grasmash: Trouble is this may break ACSF sites. Not sure those actually start with `prod:`

grasmash: Yeah, possible prefix values are devcloud, gardens, prod, network, enterprise-g1.

src/Command/CommandBase.php
```diff
@@ -1089,7 +1089,13 @@ public static function getAliasCache() {
    */
   protected function doGetApplicationFromAlias($application_alias) {
     $acquia_cloud_client = $this->cloudApiClientService->getClient();
-    $acquia_cloud_client->addQuery('filter', 'hosting=prod:' . $application_alias);
+    $possible_hosting_prefixes = ['prod', 'devcloud', 'network', 'enterprise-g1', 'gardens'];
```
anavarre, line 1092: Looks like you've already figured this out but FYI you can find the info for any sub with:

```
acli api:applications:find <UUID> | jq -r '.hosting.id' | tr -d \" | sed 's/:.*//'
```

danepowell:
If I understand correctly, the problem is that we were querying Cloud API applications using a wildcard filter, which meant that an alias of "foo" could return "openfoo"?

Instead of enumerating the hosting realms, wouldn't it be easier to change the filter to `hosting=@*:`? I tested briefly and this seems to work.

src/Command/CommandBase.php
```diff
@@ -1089,7 +1089,8 @@ public static function getAliasCache() {
    */
   protected function doGetApplicationFromAlias($application_alias) {
     $acquia_cloud_client = $this->cloudApiClientService->getClient();
-    $acquia_cloud_client->addQuery('filter', 'hosting=@*' . $application_alias);
+    $filter_value = self::createHostingIdParamValue($application_alias);
+    $acquia_cloud_client->addQuery('filter', 'hosting=' . $filter_value);
```
danepowell, line 1093:
```suggestion
    $acquia_cloud_client->addQuery('filter', 'hosting=@*:' . $filter_value);
```

Just changing this (and nothing else) seems to fix the issue for me.

grasmash:
> If I understand correctly, the problem is that we were querying Cloud API applications using a wildcard filter, which meant that an alias of "foo" could return "openfoo"?
> 
> Instead of enumerating the hosting realms, wouldn't it be easier to change the filter to `hosting=@*:`? I tested briefly and this seems to work.

🤦

PR #519, opened 2021-04-12
Fail silently when unable to find update.

grasmash:


src/Command/CommandBase.php
```diff
@@ -1192,22 +1192,26 @@ protected function checkForNewVersion(InputInterface $input, OutputInterface $ou
...
-    $response = $client->get('https://api.github.com/repos/acquia/cli/releases');
-    $releases = json_decode($response->getBody());
-
-    if (!isset($releases[0])) {
-      throw new \Exception('API error - no release found at GitHub repository acquia/cli');
+  protected function hasUpdate(): bool {
+    try {
+      $client = $this->getUpdateClient();
+      // Intentionally silence and allow errors. It's ok if this fails.
+      $response = @$client->get('https://api.github.com/repos/acquia/cli/releases');
```
anavarre, line 1201: I tried changing this URL to a one that doesn't exist to test out this PR. Didn't work despite wiping `vendor`, then doing a `ckc` and `cc`. Any idea how to try running into this issue?

src/Command/CommandBase.php
```diff
@@ -1175,11 +1175,11 @@ protected function checkForNewVersion(InputInterface $input, OutputInterface $ou
       return;
     }
     // Bale in Cloud IDEs to avoid hitting Github API rate limits.
-    if (!AcquiaDrupalEnvironmentDetector::isAhIdeEnv()) {
```
grasmash, line 1178: @anavarre 🤔 Seems important to fix.

anavarre: Without this IDEs were constantly asking to update and customers were doing it, which had us run into API limits without a few hours of an ACLI release.

grasmash: I'm not removing this line, I'm removing the `!`

grasmash: When I acli `master` in an IDE, the code does (incorrectly) progress to execute `$this->hasUpdate()`.

src/Command/CommandBase.php
```diff
@@ -1170,16 +1170,16 @@ protected function checkForNewVersion(InputInterface $input, OutputInterface $ou
...
     }
-    // Bale for development builds.
+    // Bail for development builds.
     if ($this->getApplication()->getVersion() == '@package_version@') {
       return;
     }
-    // Bale in Cloud IDEs to avoid hitting Github API rate limits.
-    if (!AcquiaDrupalEnvironmentDetector::isAhIdeEnv()) {
+    // Bail in Cloud IDEs to avoid hitting Github API rate limits.
+    if (AcquiaDrupalEnvironmentDetector::isAhIdeEnv()) {
```
danepowell, line 1178: We don't have to solve this here, but I don't like that ACLI has hardcoded support for IDEs. I'd rather support an environment variable like `ACLI_DISABLE_UPDATES`, which could be set in any environment.

src/Command/CommandBase.php
```diff
@@ -1170,16 +1170,16 @@ protected function checkForNewVersion(InputInterface $input, OutputInterface $ou
...
       return;
     }
     try {
-      if ($this->hasUpdate()) {
+      if (@$this->hasUpdate()) {
```
danepowell, line 1182: I don't feel good about this error suppression. We already have so much trouble debugging failed update checks.

Is there a way to tie this to verbosity? So we can at least see errors in verbose output.

src/Command/CommandBase.php
grasmash: ACLI has hardcoded support for IDEs in _a lot_ of places. Here we're just fixing a but, not integrating it more deeply.

src/Command/CommandBase.php
grasmash: Not that I know of. The problem is that you can't catch some php error and warnings.

danepowell (approved):

PR #518, opened 2021-04-09
Fix #495: [windows] No browser found

danepowell:
**Motivation**
Fixes #495

**Proposed changes**
On Windows, assume `start` exists instead of checking for it (AFAICT, it _always_ exists as a built-in in common shells, and there's not a consistent way of finding commands across shells on Windows).

Also improve error output to guide people towards self-solving the issue.

**Alternatives considered**
Fix `commandExists` to properly find commands on Windows. Good luck with that one, I couldn't untangle how to check that `start` exists, and any solution would probably be shell-specific.

**Testing steps**
1. Follow the [contribution guide](https://github.com/acquia/cli/blob/master/CONTRIBUTING.md#building-and-testing) to set up your development environment.
2. Clear the kernel cache to pick up new and changed commands: `acli ckc`
3. Run `auth:login` on Windows and check that it opens a browser window when prompted.

**Merge requirements**
- [x] _Bug_, _enhancement_, or _breaking change_ label applied
- [ ] Manual testing by a reviewer

PR #520, opened 2021-04-12
Upload acli.phar on every commit

danepowell:
**Motivation**
We regularly need to enlist the help of people less familiar with the ACLI codebase to test PRs or HEAD. We can facilitate this by providing a pre-built Phar for any commit (rather than forcing people to build from source).

**Proposed changes**
For any commit that runs on Travis CI (master or PRs), upload a phar to transfer.sh. We can grab a link to this phar from the Travis CI logs and distribute to anyone for testing. The link is good for 14 days.

**Testing steps**
1. Check the Travis CI logs on this PR and ensure you can download the ACLI dev build (I'll add a link here):
- https://travis-ci.com/github/acquia/cli/jobs/497913802#L502 (click to expand)
- https://transfer.sh/4wc5M/acli.phar

**Merge requirements**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
         # Warm the symfony cache so it gets bundled with phar.
         - ./bin/acli
       script: composer box-compile
-      after_script: curl --upload-file build/acli.phar https://transfer.sh/acli.phar
+      after_script: echo "$(curl --upload-file build/acli.phar https://transfer.sh/acli.phar)"
       deploy:
         provider: releases
         api_key:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ jobs:
         # Warm the symfony cache so it gets bundled with phar.
         - ./bin/acli
       script: composer box-compile
+      after_script: curl --upload-file build/acli.phar https://transfer.sh/acli.phar
       deploy:
         provider: releases
         api_key:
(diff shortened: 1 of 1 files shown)


grasmash:
Seems useful!

.travis.yml
```diff
@@ -34,6 +34,7 @@ jobs:
         # Warm the symfony cache so it gets bundled with phar.
         - ./bin/acli
       script: composer box-compile
+      after_script: echo "$(curl -s --upload-file build/acli.phar https://transfer.sh/acli.phar)"
```
grasmash, line 37: Would this happen for PR commits too?

danepowell: Yes, this will run on PR commits. Which is great for issues like #518, so the reporting user can help test.

That's also why we use transfer.sh, because it requires no auth (Travis CI can't use API keys in PRs).

anavarre: What if you commit 10 different things locally and push? Will it only create the Phar based on the last pushed commit?

danepowell: If you push multiple commits at once, it will only build on the latest commit because Travis will only run once on that commit.

This runs in Travis so it can only build when Travis runs.

anavarre (approved):
I think that's a great idea and very useful for our customers. Thanks for doing it!

Confirmed that:

- Expanding the Travis log file, I can find the transfer.sh unique link
- Clicking on the link brings me to a transfer.sh download page and the download works fine
- Checking ACLI's version locally then gave me `1.9.0-14-g78979c9` which is exactly at the SHA1 for the Travis build

PR #543, opened 2021-05-11
Warn users to configure sites.php

danepowell:
**Motivation**
Fixes DX-3908

If users fail to configure sites.php for an environment and pull a db or files from a multisite install, they will be pretty confused when Drupal loads the default site instead of whatever multisite they selected.

Note that this isn't ACLI misbehaving, it always pulls the db or files as expected. The problem is that Drupal may not know to utilize the db/files that were just pulled, causing a split-brain scenario where it tries to load files from the wrong site directory.

**Proposed changes**
Warn users of the need to configure sites.php

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -1434,6 +1434,7 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
       $site = reset($sites);
       return $site;
     }
+    $this->io->writeln("This is a multisite application. Drupal will load the default site unless you've configured sites.php for this environment: https://docs.acquia.com/cloud-platform/develop/drupal/multisite/");
     return $this->io->choice('Choose a site', $sites, $sites[0]);
   }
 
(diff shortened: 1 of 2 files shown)


src/Command/CommandBase.php
```diff
@@ -1434,6 +1434,7 @@ protected function promptChooseCloudSite($cloud_environment) {
       $site = reset($sites);
       return $site;
     }
+    $this->io->writeln("This is a multisite application. Drupal will load the default site unless you've configured sites.php for this environment: https://docs.acquia.com/cloud-platform/develop/drupal/multisite/");
```
grasmash, line 1437: We should really only write this text once. Throw it in a method and call it from both places.

grasmash (changes requested):
DRYer. :)

src/Command/CommandBase.php
grasmash: You can also consider using $this->io->warning(), but it's a bit alarming.

grasmash: Actually $this->io->note() is probably better. Looks like this:
image

grasmash (approved):

PR #544, opened 2021-05-12
Support use of refresh token for authentication

grasmash:
**Motivation**

Adds support for using a refresh token for authentication. 

**Proposed changes**

When sending requests to Cloud API, ACLI will now authenticate by:
1. Checking for either the ACLI_REFRESH_TOKEN env var to find a refresh token. If found, this will be used for authentication.
2. Fall back to looking for saved credentials in cloud_api.conf.

**Alternatives considered**

Using the access token directly. This seemed like a bad idea since League's OAuth Library already has build in support for checking access token expiry and fetching a new one by using a refresh token.

**Testing steps**

3. Get a real refresh token:
* Open Google Chrome Dev Tools
* Change to "network" tab to begin recording requests
* Login to Acquia Cloud
* Immediately, before changing URLs, filter the network tab for the "token" resource and select it
* Select the 'preview' subtab to see find the refresh token
4. Run `export ACLI_REFRESH_TOKEN=[my token]`
5. Run an Acquia CLI command and validate that authentication works, even without creds present. E.g., `./bin/acli api:applications:find` 

**Merge requirements**

### DIFF
--- a/config/prod/services.yml
+++ b/config/prod/services.yml
@@ -30,7 +30,7 @@ services:
       - ../../src/Command/Api/ApiCommandBase.php
       - ../../src/Command/Api/ApiListCommandBase.php
       - ../../src/DataStore/YamlStore.php
-      - ../../src/RefreshConnector.php
+      - ../../src/CloudApi/RefreshTokenConnector.php
     public: true
     resource: ../../src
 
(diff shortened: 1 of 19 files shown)


danepowell (approved):

anavarre:
Would like to request @itafroma 's review here.

grasmash:
It appears that this won't work. ACLI would ALSO need both the client id and secret in order to refresh the access token, but it does not have access to those things in the IDE.

Therefore we should have the proxy container simply pass a valid access token to ACLI and the proxy should handle the refresh logic.

PR #545, opened 2021-05-13
Use ACLI_ACCESS_TOKEN for auth if present

grasmash:
**Motivation**

Adds support for using access token defined in ACLI_ACCESS_TOKEN.

**Proposed changes**

Before looking for a saved token in cloud_api.conf, ACLI will first check to see if $ACLI_ACCESS_TOKEN is defined as an env var. If so, ACLI assumes that it is a valid access token (not expired) and will use it for authentication.

**Alternatives considered**

Previously considered getting a refresh token from an env var, but that would have required knowing the client id and other inaccessible information.

**Testing steps**

3. Get a real access token. It will expire after 5 minutes
* Open Google Chrome Dev Tools
* Change to "network" tab to begin recording requests
* Login to Acquia Cloud
* Immediately, before changing URLs, filter the network tab for the "token" resource and select it
* Select the 'preview' subtab to see find the access token
4. Run `export ACLI_ACCESS_TOKEN=[my token]`
5. If you have a local ~/.acquia/cloud_api.conf, run `mv ~/.acquia/cloud_api.conf ~/.acquia/cloud_api.conf.bak` so that it does not interfere
6. Run an Acquia CLI command and validate that authentication works, even without creds present. E.g., `./bin/acli api:applications:find` 
7. When done, restore your local creds file with `mv ~/.acquia/cloud_api.conf.bak ~/.acquia/cloud_api.conf` 

**Merge requirements**


danepowell (approved):

sumitmotghare:
The Happy Path work's here but the below use case fails.

Screenshot 2021-05-14 at 2 28 16 PM

If the token is expired and user authenticates himself/herself with the `key` and `secret` then it fails the retrieve the details and throw's the same error as `expired token`

Screenshot 2021-05-14 at 2 33 19 PM

User need to explicitly unset the environment variable to access `acli` via `key` and `secret`.

Screenshot 2021-05-14 at 2 37 18 PM

anavarre:
@sumitmotghare please remove / blur any sensitive information like UUIDs or else. Acquia CLI is a public repository and we don't want any sensitive information to leak out.

grasmash:
@sumitmotghare do you expect the env var to ever contain an expired key?

sumitmotghare:
> @sumitmotghare do you expect the env var to ever contain an expired key?

It should not, but for fallback we should have provision to avail user a benefit of having `key` and `secret` incase `auth_token` didn't worked.
But I am ok, if we don't want to entertain any of the fallback's.

grasmash:
I don’t love the idea of making two requests in the case of an invalid access token. However, it would be pretty simple for Cloud IDE to set an additional environmental variable indicating the access token expiration time. We could check that before deciding to use the access token. That would obviate the need for two requests.

PR #591, opened 2021-06-09
Change IDE SSH key generation

sumitmotghare:
**Motivation**

Fixes the broken GIT UI for Cloud IDE

**Proposed changes**

- Temporarily support SSH key generation without passphrase key
- Static file name for all SSH keys in the IDE. [ Dynamic in Cloud Platform ]

**Alternatives considered**

- File support ticket on Theia to consider passphrase keys for Git UI

**Testing steps**

- Apply this PR on any of the existing IDE.
- Remove the existing Keys for the IDE from ~/.ssh, ssh-add -D and Cloud platform.
- Authenticate the IDE against `acli`
- Create a new Key using `{path to acli}/bin/acli ide:wizard:ssh-key:create-upload`
- Clone the Code from Cloud Platform.
- Try to use Git UI in the IDE to pull/push the committed code.

**Merge requirements**
- [X] Manual testing by a reviewer

### DIFF
--- a/src/Command/Ssh/SshKeyCreateCommand.php
+++ b/src/Command/Ssh/SshKeyCreateCommand.php
@@ -27,7 +27,8 @@ class SshKeyCreateCommand extends SshKeyCommandBase {
   protected function configure() {
     $this->setDescription('Create an SSH key on your local machine')
       ->addOption('filename', NULL, InputOption::VALUE_REQUIRED, 'The filename of the SSH key')
-      ->addOption('password', NULL, InputOption::VALUE_REQUIRED, 'The password for the SSH key');
+      ->addOption('password', NULL, InputOption::VALUE_REQUIRED, 'The password for the SSH key')
+      ->addOption('is-wizard', FALSE, InputOption::VALUE_REQUIRED, 'The flag for IDE wizard');
   }
 
   /**
@@ -128,13 +129,14 @@ class SshKeyCreateCommand extends SshKeyCommandBase {
    * @throws \Exception
    */
   protected function determinePassword(InputInterface $input, OutputInterface $output): string {
-    $password = "";
-    if ($input->getOption('password')) {
-      $password = $input->getOption('password');
-      $this->validatePassword($password);
+    if ($input->getOption('is-wizard')) {
+      $password = "";
     }
     else {
-      if ($input->getOption('password') !== "") {
+      if ($input->getOption('password')) {
+        $password = $input->getOption('password');
+        $this->validatePassword($password);
+      } else {
         $question = new Question('Enter a password for your SSH key');
         $question->setHidden($this->localMachineHelper->useTty());
         $question->setNormalizer(static function ($value) {
@@ -155,7 +157,8 @@ class SshKeyCreateCommand extends SshKeyCommandBase {
    */
   protected function validatePassword($password) {
     $violations = Validation::createValidator()->validate($password, [
-      new Length(['min' => 0])
+      new Length(['min' => 5]),
+      new NotBlank(),
     ]);
     if (count($violations)) {
       throw new ValidatorException($violations->get(0)->getMessage());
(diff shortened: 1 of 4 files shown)


anavarre (commented):
Note: PR would still offer a default value which will be `id_rsa_acquia`.

sumitmotghare:
> Note: PR would still offer a default value which will be `id_rsa_acquia`.

For command `{path to acli}/bin/acli ide:wizard:ssh-key:create-upload` it will always offer an `id_rsa` value. 

- The `if` condition will always satisfy for the above command as the parent function returns the required value. 
- `else` will triggered only when the filename is not passed. 
 
To reproduce the issue that you mentioned, user need to execute command `{path to acli}/cli/bin/acli ssh-key:create` which will ask for the filename. And if the filename not passed in that case it will take the default file name i.e. `id_rsa_acquia`

anavarre:
> And if the filename not passed in that case it will take the default file name i.e. id_rsa_acquia

What I'm saying is when we don't give a default value, we want it to be `id_rsa`, since we know anything else will break.

sumitmotghare:
> > And if the filename not passed in that case it will take the default file name i.e. id_rsa_acquia
> 
> What I'm saying is when we don't give a default value, we want it to be `id_rsa`, since we know anything else will break.

Addressed here - https://github.com/acquia/cli/pull/591/commits/494d1f2ae4b9f8834c2a158200f97e93f1c9975a

anavarre:
So, this works better now:

```
$ /tmp/cli/bin/acli ssh-key:create-upload

 Please enter a filename for your new local SSH key. Press enter to use default value [id_rsa]:
 > 

 Enter a password for your SSH key:
 > 


 Please enter a Cloud Platform label for this SSH key:
 > brokengitui

Uploaded id_rsa.pub to the Cloud Platform with label brokengitui
    ✔ Waiting for SSH key to become available on the Cloud Platform...

Your SSH key is ready for use.
```

And

```
$ /tmp/cli/bin/acli ide:wizard:ssh-key:create-upload
    ✔ Creating a local SSH key
    ✔ Uploading the local key to the Cloud Platform
    ✔ Adding the SSH key to local keychain
    ✔ Waiting for the key to become available on the Cloud Platform

Your SSH key is ready for use!

$ ssh-add -l
4096 SHA256:ex876bCsOfqpZAe86DRxeB2Wwr3h2vGnOemvLKFFEeE /home/ide/.ssh/id_rsa (RSA)
$ ssh-add -D
All identities removed.
$ keychain ~/.ssh/id_rsa

 * keychain 2.8.1 ~ http://www.funtoo.org
 * Found existing ssh-agent: 17895
 * Adding 1 ssh key(s): /home/ide/.ssh/id_rsa
 * ssh-add: Identities added: /home/ide/.ssh/id_rsa

$ ssh-add -l
4096 SHA256:ex876bCsOfqpZAe86DRxeB2Wwr3h2vGnOemvLKFFEeE /home/ide/.ssh/id_rsa (RSA)
```

src/Command/Ssh/SshKeyCreateCommand.php
```diff
@@ -152,8 +155,7 @@ protected function determinePassword(InputInterface $input, OutputInterface $out
    */
   protected function validatePassword($password) {
     $violations = Validation::createValidator()->validate($password, [
-      new Length(['min' => 5]),
-      new NotBlank(),
+      new Length(['min' => 0])
```
grasmash, line 158: I would prefer to leave this validation here and conditionally bypass it when inside of an IDE

grasmash: Or better yet, when we know that we are calling it from the wizard command.

src/Command/Ssh/SshKeyCreateCommand.php
```diff
@@ -128,18 +128,21 @@ protected function validateFilename($filename) {
    * @throws \Exception
    */
   protected function determinePassword(InputInterface $input, OutputInterface $output): string {
+    $password = "";
     if ($input->getOption('password')) {
       $password = $input->getOption('password');
       $this->validatePassword($password);
     }
     else {
-      $question = new Question('Enter a password for your SSH key');
```
grasmash, line 136: We still want to prompt people for a password when they are doing this on their local machine.

sumitmotghare: Now it will prompt for the password if user tries to create the SSH key for local machine.

PR #607, opened 2021-07-19
Fix PHPCS

danepowell:
**Motivation**
PHPCS is not catching errors in Travis CI

**Proposed changes**
Remove the problematic exclude-pattern (see https://github.com/squizlabs/PHP_CodeSniffer/issues/981 )

**Testing steps**
See that the first commit fails because it properly enables PHPCS, then the second commit passes after fixing the style issues.

**Merge requirements**

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -12,7 +12,9 @@
 
   <file>.</file>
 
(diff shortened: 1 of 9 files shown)


grasmash:
Tricksy hobbitses.

PR #634, opened 2021-09-01
Fixes #530: Ensure Git and Composer exist

### LINKED ISSUE #530: [push:artifact] ensure Git and Composer exist
`push:artifact` relies on Git and Composer existing, and should abort if they do not. We should decide on one or more solutions:
- Follow the pattern used elsewhere and just call `LocalMachineHelper::commandExists()` at the start of the script. This is okay-ish, but I'm not sure it scales as more commands shell out.
- Modify `LocalMachineHelper::execute()` to check for the existence of _any_ command it runs, so we don't have to manually verify each one. This scales much more nicely, though individual commands lose control of when this validation occurs. We could implement this even if we keep manual validation in some places, my only concern is that it won't find aliases or shell built-ins, which could lead to problems such as https://github.com/acquia/cli/issues/495
- Handle this as part of https://github.com/acquia/cli/issues/525. This is a sort of hybrid solution... it at least scales for Git and Composer, but has fewer potential side effects than universally checking commands in `LocalMachineHelper::execute()`.

grasmash:
**Motivation**

Fixes #530: Ensure Git and Composer exist

**Proposed changes**

Before executing any process, validate that the required binaries exist on the local machine. If they do not, throw an error. 

**Alternatives considered**

**Testing steps**

**Merge requirements**


src/Helpers/LocalMachineHelper.php
```diff
@@ -45,9 +47,26 @@ public function __construct(InputInterface $input, OutputInterface $output) {
    * @return bool
    */
   public function commandExists($command): bool {
+    if (array_key_exists($command, $this->installedBinaries)) {
+      return (bool) $this->installedBinaries[$command];
+    }
     $os_command = OsInfo::isWindows() ? ['where', $command] : ['which', $command];
     // phpcs:ignore
-    return $this->execute($os_command, NULL, NULL, FALSE)->isSuccessful();
+    $exists = $this->execute($os_command, NULL, NULL, FALSE)->isSuccessful();
```
danepowell, line 55: @grasmash this only checks if the command exists in the PATH. It will fail in two scenarios, which are common enough that I think we need to account for:

1. A command that exists relative to the current working directory, i.e. `vendor/bin/foo`. This should be easy to check for, just see if the command exists relative to the cwd.
2. A command that's defined as an alias, i.e. `alias composer='composer2'`. Use `command -v` instead of `which` to solve this. [reference](https://stackoverflow.com/questions/592620/how-can-i-check-if-a-program-exists-from-a-bash-script)

None of this accounts for Windows (I'm not enough of an expert to say), and there are also some edge cases it may miss (if a command exists but isn't executable). I'm fine with burying our head in the sand on those.

These are all problems we faced previously on ADS CLI and BLT, FWIW. Makes me think PHP needs a common library to deal with this kind of thing properly.

grasmash: > A command that exists relative to the current working directory, i.e. vendor/bin/foo. This should be easy to check for, just see if the command exists relative to the cwd.

Even if it does exist, we'd need to actually call it using the directory path. That would require a lot of changes to many commands in ACLI. Seems like a separate PR to me.

danepowell (approved):

PR #640, opened 2021-09-02
High memory usage during pull:db

danepowell:
**Motivation**
Running pull:db with a big db dump can exhaust available memory.

**Proposed changes**
Stream db backup contents to disk to limit memory usage to the stream buffer size (<22MB in my testing). See https://github.com/typhonius/acquia-php-sdk-v2/issues/185

Getting tests working across OSes required significant refactoring of mocking methods, but I think it's for the better (I was able to delete a static fixture).

**Testing steps**
3. Use an environment with limited memory, or limit the availabe memory via php.ini / cli options (`php -d memory_limit=50M bin/acli...`)
3. `wacli pull:db <big-environment-id>`

**Merge requirements**


src/Command/Api/ApiCommandBase.php
```diff
@@ -91,6 +91,10 @@ protected function execute(InputInterface $input, OutputInterface $output) {
     }
 
     $path = $this->getRequestPath($input);
+    $acquia_cloud_client->addOption('headers', [
+      'Accept' => 'application/json',
```
danepowell, line 95: I don't know if this will work as expected, it might clobber the existing `User-Agent` header.

danepowell:
@grasmash why did Labeler act on this PR? Also, why is Travis CI not running? I assume related to the recent Actions changes?

src/Command/Api/ApiCommandBase.php
danepowell: I confirmed via server logs that the user agent is still applied as before.

src/CloudApi/ClientService.php
```diff
@@ -42,7 +42,6 @@ public function getClient(): Client {
     $user_agent = sprintf("acli/%s", $this->application->getVersion());
     $client->addOption('headers', [
       'User-Agent' => [$user_agent],
-      'Accept'     => 'application/json',
```
danepowell, line 45: This forces all API calls (even internal ones) to return JSON. That's obviously bad if what we want from the API is a DB backup as an octet stream. So I moved this to only be set on public API calls.

src/Command/Pull/PullCommandBase.php
```diff
@@ -47,6 +47,24 @@ abstract class PullCommandBase extends CommandBase {
...
   protected $drushHasActiveDatabaseConnection;
 
+  /**
+   * @param $environment
+   * @param $database
+   * @param $backup_response
+   *
+   * @return string
+   */
+  public static function getBackupPath($environment, $database, $backup_response): string {
```
danepowell, line 57: Extracted this method so that it can be used in tests to get a consistent path across OSes.

grasmash (approved):

PR #646, opened 2021-09-10
Migrate from Travis CI to GitHub Actions

danepowell:
This is working and ready for review. `softprops/action-gh-release@v1` is approved and enabled on the org.

Until this is merged, it won't actually run on the Acquia account. You can see test results for this feature branch on my fork: https://github.com/danepowell/cli/actions

You can also see a tagged release where Actions attached the phar: https://github.com/danepowell/cli/releases/tag/actions1


grasmash:
Do we really need to move since this is OSS?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
+name: CI
+on:
+  push:
```
grasmash, line 3: are the trailing colons necessary here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    env:
+      ACLI_PRINT_COMMAND_OUTPUT: 1
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+        php: [ "7.3", "7.4", "8.0"]
+        coverage: [ none ]
+        include:
```
grasmash, line 23: Not urgent, but this seems like it's running one more job than necessary. Do we need a dedicated job for this, or can me run it as part of one of the other builds?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,73 @@
...
+          # Generate .env.local.php
+          composer dump-env prod
+          # Warm the symfony cache so it gets bundled with phar.
+          ./bin/acli
+          composer box-compile
+          echo "$(curl -s --upload-file build/acli.phar https://transfer.sh/acli.phar)"
+      - name: Release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
```
grasmash, line 72: Just curious if you've tested this works?

danepowell:
> Do we really need to move since this is OSS?

Yes. You'll see the ticket in the backlog shortly with more details, but this is mandatory.

.github/workflows/ci.yml
danepowell: Trailing colon is required, since it's a mapping, not a simple list/array. [Docs explicitly say so](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration)

> You must append a colon (:) to all events, including events without configuration

.github/workflows/ci.yml
danepowell: Took some digging, but I was able to eliminate this job without embiggening the build file, taking inspiration from https://github.com/actions/runner/issues/409

.github/workflows/ci.yml
danepowell: Yep, see the link in the description, I created a test release on my fork with phar attached via this action: https://github.com/danepowell/cli/releases/tag/actions1

danepowell:
@grasmash The upstream action was approved, so this can be merged whenever you're comfortable with it.

PR #648, opened 2021-09-15
Fixes #641: Trigger custom scripts after acli pull commands

### LINKED ISSUE #641: Trigger custom scripts after `acli pull` commands
### Story

As a user, I'd like to run things like `drush db:update` automatically each time that I run `acli pull:db`.

### Implementation
I'd like to avoid introducing a new event system. However, I'm open to piggy-backing on an existing one, like Composer's. I propose that we check for the existence of a `post-acli-pull-db` script in composer.json, and if it exists, run `composer post-acli-pull-db`.

grasmash:
**Motivation**

Fixes #641

**Proposed changes**

After any ACLI command is run, ACLI will check `$(pwd)/composer.json` for a `post-acli-($command name)` script and execute it if present. E.g., `post-acli-db:pull`.

**Alternatives considered**

**Testing steps**

3. `cd` into a Drupal application directory containing a composer.json.
4. Add an entry to the `scripts` key in the composer.json:
```
        "post-acli-ide-list": [
            "echo \"Hello world\""
        ]
```
5. Run `[path to acli]/bin/acli ide-list -vvv`
6. Observe:
```
Executing composer script `post-acli-ide-list` defined in `/Users/matt.grasmick/Sites/projects/amad9/composer.json`
> echo "Hello world"
Hello world
[notice] Command: 'composer' 'run-script' 'post-acli-ide-list' [Exit: 0]
```

**Merge requirements**
- [x] Manual testing by a reviewer

### DIFF
--- a/bin/acli
+++ b/bin/acli
@@ -20,6 +20,8 @@ use Acquia\Cli\Command\Api\ApiCommandHelper;
 use Acquia\Cli\Helpers\LocalMachineHelper;
 use SelfUpdate\SelfUpdateCommand;
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Console\Output\ConsoleOutput;
+use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
 use Symfony\Component\Filesystem\Filesystem;
@@ -58,7 +60,6 @@ if ($_SERVER['APP_DEBUG']) {
 if (!getenv('HOME')) {
     putenv('HOME=' . LocalMachineHelper::getHomeDir());
 }
-
 $kernel = new Kernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
 
 // Handle a clear-kernel-cache pseudo command. This isn't implemented as a true console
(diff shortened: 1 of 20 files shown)


mikemadison13:
I was able to test this locally and it seems to work!

grasmash:
Cool. 

I have no idea how to write an automated test for this. Our tests don't seem to fire the ExceptionListener class.

danepowell:
I'm onboard with the general idea of a hook system that can call out to Composer scripts, but I'm a little uncomfortable with this implementation. It's clever, but a little magical in some way that I don't like.

For instance... what if someone defines a post-pull-db script expecting it to run every time a db is imported (i.e. `pull:db`). They'd probably expect it to run on `pull` as well, right? But I don't think it will with the current implementation.

And I just imagine people demanding a more extensive hook system anyway at some point (with pre-, post-, and during hooks). Maybe it's better to consider that now.

If this implementation totally satisfies @mikemadison13 's use case, then I guess I'm cool with it. I appreciate that there's a ton of validation of the Composer scripts, so at least it's unlikely to cause problems.

grasmash:
I considered the issue of pull not extending to pull:db, but that’s something that’s easily solved by the script implementation. Composer scripts can call one another. You could create a script and just ensure that it’s called in multiple hooks.

it doesn’t seem worth adding the complexity to our CLI when a very simple implementation change on the customer end would satisfy the requirements.

I would prefer to keep this as lightweight as possible, although I would be open to adding pre-hooks as well. That’s pretty straightforward And could probably use the same logic.

mikemadison13:
i think for me to really say definitively the next step would be to start writing some scripts and doing side by side comparison of "if you run this blt command..." and then try to replicate it on this side using acli and whatever composer script. i'm sure there will be some deltas (good call out @danepowell !) but it's my hope we can work around them or evolve what ACLI does to adapt?

grasmash:
This now supports "pre-acli*" and "post-acli*" hooks.

danepowell (approved):

PR #706, opened 2021-10-29
Fixes #687: ssh-key:create ssh-agent integration and improved polling

### LINKED ISSUE #687: ssh-key:create should add key to local ssh-agent

grasmash:
**Motivation**

Fixes #687

**Proposed changes**

* Add the new ssh key to the local ssh-agent
* Rather than polling cloud api for the key, try to actually ssh into a web to test that the ssh key is ready
* Mock the filesystem in tests rather than making real ssh keys during testing

**Alternatives considered**

**Testing steps**

3. Validate that `ssh-key:create-upload` still works and now adds the ssh key to the local ssh agent
4. Validate that `ide:wizard` still works

**Merge requirements**


grasmash:
This was a very frustrating issue to work on. Lots of testing modification required, but hopefully we came out better for it. I suspect that getting a testing coverage up from where it is right now will be very challenging. I spent a good number of hours making it what it is.

src/Command/Ssh/SshKeyCommandBase.php
```diff
@@ -31,15 +66,149 @@ public static function getIdeSshKeyLabel(IdeResponse $ide): string {
...
+  }
+
+  /**
+   * Get the first Cloud application available.
+   *
+   * @return false|mixed|string|null
+   * @throws \Exception
+   */
+  protected function getAnyAhApplication() {
+    if ($app_uuid = $this->determineCloudApplication(FALSE, FALSE)) {
```
grasmash, line 187: Not sure if this is the right decision. If they've already got an application linked to the current repo, we check the envs for that app for ssh connectivity. Otherwise, we just grab the first application they have access to and check that. We never ask them which app to test.

@anavarre @danepowell thoughts?

src/EventListener/ExceptionListener.php
```diff
@@ -68,6 +68,7 @@ public function onConsoleError(ConsoleErrorEvent $event): void {
       switch ($errorMessage) {
         case "There are no available Cloud IDEs for this application.\n":
           $this->helpMessages[] = "Delete an existing IDE via <bg={$this->messagesBgColor};fg={$this->messagesFgColor};options=bold>acli ide:delete</> or contact your Account Manager or Acquia Sales to purchase additional IDEs.";
+          // @todo
```
danepowell, line 71: Stray todo?

danepowell:
> Rather than polling cloud api for the key, try to actually ssh into a web to test that the ssh key is ready

What's the purpose of this? Is the API not reliable? Can we tell the difference between an ssh connection failing because the ssh key is not yet uploaded versus failing for other technical reasons?

src/Command/Api/ApiCommandHelper.php
```diff
@@ -547,8 +547,6 @@ protected function getSkippedApiCommands(): array {
       'accounts:drush-aliases',
       // Skip any command that has a duplicative corresponding ACLI command.
       'ide:create',
-      'ide:delete',
```
danepowell, line 550: This seems like an unrelated (and undesirable) change?

src/Command/CommandBase.php
```diff
@@ -857,12 +858,12 @@ protected function doDetermineCloudApplication() {
     }
 
     // Try to guess based on local git url config.
-    if ($cloud_application = $this->inferCloudAppFromLocalGitConfig($acquia_cloud_client)) {
+    if ($prompt && $cloud_application = $this->inferCloudAppFromLocalGitConfig($acquia_cloud_client)) {
```
danepowell, line 861: Reading from local Git config doesn't involve a prompt, does it? So is there a reason to skip it if `$prompt == FALSE`?

danepowell: Oh I see, it's because we're overloading this function to "get any app at all". If we decide to keep this approach, maybe there's a better name for the `$prompt` variable.

src/Command/Ssh/SshKeyCommandBase.php
danepowell: I think we should force selecting/linking to a specific application to run this command.

A customer can have application access (i.e. UI / API access) without having SSH access. For this to work without user input, we'd have to separately query the permissions API to figure out which applications allow SSH access. Even then, it's not ideal since different servers might install the SSH keys at different times. Just because one works doesn't mean the others will.

danepowell (changes requested):

src/Command/Ssh/SshKeyCommandBase.php
grasmash: Ah.

src/Command/Api/ApiCommandHelper.php
grasmash: Unrelated, yes. Undesirable, no.

grasmash:
> What's the purpose of this? Is the API not reliable?

Correct. The API is not reliable for knowing that SSH access is actually ready.

src/Command/Ssh/SshKeyCommandBase.php
grasmash: Trouble with this is that the user may be confused about why we're asking them to select an application.

danepowell: Let's just be transparent, we need them to link the app so we can verify that the SSH key was installed correctly. We can make it optional: if they don't want to link the app, we won't verify the installation.

src/Command/CommandBase.php
```diff
@@ -805,18 +805,19 @@ protected function determineCloudEnvironment() {
...
-   * @param bool $link_app
+   * @param bool $prompt_link_app
+   * @param bool $prompt_app
    *
    * @return string|null
    * @throws \Exception
    */
-  protected function determineCloudApplication($link_app = FALSE): ?string {
-    $application_uuid = $this->doDetermineCloudApplication();
+  protected function determineCloudApplication($prompt_link_app = FALSE, $prompt_app = TRUE): ?string {
```
danepowell, line 814: `$prompt_app` argument appears to never be passed and can be removed. Same for `doDetermineCloudApplication`.

src/Command/CommandBase.php
```diff
@@ -857,12 +858,12 @@ protected function doDetermineCloudApplication() {
     }
 
     // Try to guess based on local git url config.
-    if ($cloud_application = $this->inferCloudAppFromLocalGitConfig($acquia_cloud_client)) {
+    if ($prompt_user && $cloud_application = $this->inferCloudAppFromLocalGitConfig($acquia_cloud_client)) {
```
danepowell, line 861: Same as previous comment but just to reiterate, `$prompt_user` appears to always be TRUE and can be removed.

src/Command/Ssh/SshKeyCommandBase.php
```diff
@@ -31,15 +66,131 @@ public static function getIdeSshKeyLabel(IdeResponse $ide): string {
...
+  protected function pollAcquiaCloudUntilSshSuccess(
+    OutputInterface $output
+  ): void {
+    // Create a loop to periodically poll the Cloud Platform.
+    $loop = Loop::get();
+    $spinner = LoopHelper::addSpinnerToLoop($loop, 'Waiting for the key to become available on the Cloud Platform', $output);
+
+    // Wait for SSH key to be available on a web.
+    $cloud_app_uuid = $this->determineCloudApplication(TRUE);
+    $environment = $this->getAnyAhEnvironment($cloud_app_uuid);
```
danepowell, line 157: Sorry, maybe I should have been more specific before, because the same problem of checking "any" application applies to checking "any" environment for a given app. Many users will have SSH access to dev/stage, but not prod. This check will thus time out for many developers.

Let's modify `getAnyAhEnvironment()` to prefer dev/stage. It's a heuristic check but seems safe, I can't imagine many people have access to prod but _not_ dev/stage. The alternative is asking people to select a specific environment (confusing UX), or polling the permissions API to find an environment with SSH access (complex).

src/Command/Ssh/SshKeyUploadCommand.php
```diff
@@ -58,13 +59,40 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 
     // Wait for the key to register on the Cloud Platform.
     if ($input->getOption('no-wait') === FALSE) {
-      $this->output->write('Waiting for new key to be provisioned on the Cloud Platform...');
-      $this->pollAcquiaCloud($output, $acquia_cloud_client, $public_key);
+      if ($this->input->isInteractive()) {
+        $this->io->note("It may take some time before the SSH key is installed on all of your application's web servers.");
+        $answer = $this->io->confirm("Would you like to wait until Cloud Platform is ready?");
+        if (!$answer) {
+          $this->io->success('Your SSH key cas been successfully uploaded to Cloud Platform.');
```
danepowell, line 66:
```suggestion
          $this->io->success('Your SSH key has been successfully uploaded to Cloud Platform.');
```

danepowell (changes requested):

danepowell:
I edited the title to better reflect the scope of this PR, since we're overloading it with two issues it seems (ssh-agent integration and improved polling). We might want to split this up, since ssh-agent integration seems pretty well solved while SSH polling is turning into a pretty kettle of fish.

danepowell (approved):
LGTM. Unfortunately I can't test manually as the recent Cloud password changes somehow locked me out of my account.

grasmash:
image
image

grasmash:
image

PR #719, opened 2021-11-10
Fixes CLI-141: Support federated sign-on.

grasmash:
- Fixes CLI-141: Support federated sign-on.
- Tested.
- Grab from AH_ORGANIZATION_UUID.

**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**


src/Command/CommandBase.php
```diff
@@ -408,6 +414,31 @@ public function checkAndPromptTelemetryPreference(): void {
...
+   *
+   * This is a requirement for Acquia Cloud organizations that have SSO enabled.
+   *
+   * @throws \League\OAuth2\Client\Provider\Exception\IdentityProviderException
+   */
+  protected function addOrgScopeIfRequired() {
+    // AH_APPLICATION_UUID will be set in Acquia environments, including Cloud IDE.
+    if ($application_uuid = AcquiaDrupalEnvironmentDetector::getAhApplicationUuid()) {
+      try {
+        $application = $this->getCloudApplication($application_uuid, TRUE);
```
danepowell, line 428: I don't think there's any harm in adding the org scope even if it's not required for a particular call.

Should we just check for AH_ORGANIZATION_UUID and avoid the application API call?

grasmash: Is it true that’s there’s no harm in adding it? I don’t know the implications.

danepowell: @itafroma if an organization uses single sign-on, is there ever a reason _not_ to request the `organization:uuid` scope when generating an API token? i.e. could that scope ever impair access to an API resource?

We should update the docs with the answer to this question: https://docs.acquia.com/cloud-platform/develop/api/auth/#making-api-calls-through-single-sign-on

grasmash: The environmental variable will be set even if the customer doesn’t use single sign-on.

danepowell: > The environmental variable will be set even if the customer doesn’t use single sign-on

Ah, you are right. I think it's still resolving the answer to that question.

grasmash: Based on discussion with @itafroma: don't add it if you don't need it.

>  generally clients should only request as many scopes as they need for that particular session and follow a privilege elevation model

danepowell: Something about this rubs me the wrong way. Maybe it's because I don't understand the reasoning behind requiring an organization scope in the first place.

> generally clients should only request as many scopes as they need for that particular session and follow a privilege elevation model

This implies that human involvement is somehow a part of this security model. I mean, from a technical standpoint, there's no reason that an organization scope is required: Cloud API itself could "figure out" whether an org scope is required for a given request and automatically add it.

If that's the case, then do we really want Acquia CLI to play the role of the human and automatically add the scope?

Does what I'm asking here make any sense to anyone else? 😁

anavarre: I agree with how Dane reads it. The privilege escalation model can be e.g. prompting the user for every single ACLI call where it's needed or granting access once and for all in a config file.

The way we were thinking about solving this for Cloud IDE was to automatically detect and pass on the `AH_ORGANIZATION_UUID` environment variable so that each and every ACLI request would be able to detect it, but that defeats the purpose of the privilege escalation model.

grasmash: Ok. So how should we change the approach? This PR is working right now. It doesn't require human interaction. It determines if the scope is needed by attempting to make a request and checking if it failed. If it does fail, the scope is added.

I'm not really sure what alternative is being proposed here. Do you _want_ some kind of human interaction? What question would be ask? I suppose we could write something like.

If AH_ORGANIZATION_UUID is set:

> It looks like Federated Authentication is required for this application. You must scope this ACLI request to the "Something" organization? Proceed Y/N [yes]
> Organization uuid written to .acquia-cli.yml.

If AH_ORGANIZATION_UUID is not set:

> It looks like Federated Authentication is required for this application. You must scope this ACLI request to the "Something" organization? Proceed Y/N [yes]
> Enter the uuid of the organization to which this application belongs:
> Organization uuid written to .acquia-cli.yml.

Though honestly this is kind of confusing prompt.

anavarre: # Option 1

Seems like a sub-optimal solution if a) we need to make an API call for every single request and b) it doesn't respect the idea that we should ask for the privilege escalation specifically.

# Option 2

Seems like we'd be making one API call to check the first time and store to a file? I like that.

> Prompt the user for the uuid if it's not set in an environmental variable.

You could argue you don't even need the extra complexity in Cloud IDE and would just check and write to a file when needed. This would keep things simple and would prevent you from having to do this dance with the user prompts.

# Option 3

While I like it in theory, the day we improve our Cloud IDE UI integration, we might regret not having refresh tokens available for app that require Fed Auth.

danepowell: Isn't there another option as suggested by Mark, which is to require the user to explicitly ask for the org scope whenever they need to access a protected endpoint? This seems the most proper from a security standpoint, even if it's more friction for the user.

I'd suggest catching the Cloud API error when the org scope is required and rewriting the exception to clearly indicate that the user needs to pass an org scope. Then we do one or more of the following to allow the user to _manually_ request org scope:
- Accept the org uuid via an `--org-id` argument
- Add a `--request-org-scope` argument that prompts the user for org id
- Optionally cache whatever org id was passed so the user doesn't have to remember it, _but_ still only apply the scope to a given request if `--request-org-scope` is passed.

grasmash: That only works if they're hitting a single API endpoint. Many of ACLI's commands are automations and wrappers for multiple commands.

For instance, when a user clicks "Configure IDE" in a Cloud IDE, it runs `ide:wizard`. We don't want that to fail with an exception asking them to pass the organization scope.

Similarly, do you really want users to run `acli pull --request-org-scope` ? I doubt they'll even understand what that means.

anavarre: > Similarly, do you really want users to run acli pull --request-org-scope ? I doubt they'll even understand what that means.

I'm not the target audience, but I sure don't.

itafroma: I apologize if I gave the wrong signal here: I expected option 1 to be implemented and would satisfy the privilege escalation model we use with Cloud UI. The intentionality of the customer's desire to elevate is inherent in the fact that they requested the operation for that protected resource to be performed. 

Maybe it makes sense, especially if there's a verbose mode, to add a warning to the console output like:

> Warning: this action requires access to a resource protected by Federated Authentication. Requesting access to the resource from organization <UUID>...

grasmash: @anavarre regarding your commentary on Option 1:
> Seems like a sub-optimal solution if a) we need to make an API call for every single request and b) it doesn't respect the idea that we should ask for the privilege elevation specifically.

A) we cache the API call response so it's not made on every request. It's made once per minute max.
B) It fits the intended model that @itafroma described.

anavarre: If Mark is happy with option 1, then I surrender to it 😆

grasmash: @danepowell you good?

danepowell: Yeah option 1 is good then

PR #725, opened 2021-11-19
CLI-141: Provide workaround for Fed Auth, prefer api key and secret.

grasmash:
**Motivation**

Fixes CLI-141

**Proposed changes**

1. When Federated Login is required and the user runs an acli command using the Cloud IDE provided access token, throw a useful error instructing them to run `acli login.`
2. When an API key and secret are present on disk, prefer them over the Cloud IDE defined access token.

**Alternatives considered**

**Testing steps**

**Merge requirements**


grasmash:
So, I realized that the previous implementation of this will not work. We can't use the Cloud IDE access token at all because it was never scoped to the the org. Therefore, we must force the customer to manually login via `acli` and use an API token and secret for authentication.

grasmash:
image
image

PR #737, opened 2021-12-08
Fixes #557: Throw error if $HOME cannot be determined.

grasmash:
**Motivation**

Fixes IDE-557

**Proposed changes**

If $HOME is not set and cannot be determined, throw an error.

**Merge requirements**

### DIFF
--- a/src/Helpers/LocalMachineHelper.php
+++ b/src/Helpers/LocalMachineHelper.php
@@ -286,7 +286,7 @@ class LocalMachineHelper {
     }
 
     if (!$home) {
-      throw new AcquiaCliException('Set $HOME in your shell.');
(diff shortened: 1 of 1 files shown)


src/Helpers/LocalMachineHelper.php
```diff
@@ -284,6 +285,10 @@ public static function getHomeDir(): string {
       }
     }
 
+    if (!$home) {
+      throw new AcquiaCliException('Set $HOME in your shell.');
```
danepowell, line 289:
```suggestion
      throw new AcquiaCliException('Could not determine $HOME directory. Ensure $HOME is set in your shell.');
```

anavarre:
@anujkaushal FYI this will support the work you're doing with retrieving `AH_SITE_GROUP` in Cloud IDE.

PR #739, opened 2021-12-09
Fixes #731: Compress files for push:files and pull:files commands

### LINKED ISSUE #731: Can't sync files from Prod
```
acli pull:files
```

returns 

> Unable to sync files from Cloud. could not make way for new symlink: files

I can confirm that files is a symlink, 

```
lrwxrwxrwx. 1 root        root     78 Nov 19 21:43 files -> /mnt/data/[uuid]-real-shared/sites/default/files
```

It executes,

```
rsync -rltDvPhe ssh -o StrictHostKeyChecking=no SITE.prod@SITE.ssh.prod.acquia-sites.com/mnt/gfs/SITE.prod/sites/default/files /app/docroot/sites/default/
```

Which fails. 

If I do, 

````
drush rsync @prod:/var/www/html/docroot/sites/default/files/ /app/docroot/sites/default/files/
````

it runs, 

```
rsync -e ssh -akz SITE.prod@SITE.ssh.prod.acquia-sites.com:/var/www/html/docroot/sites/default/files/ /app/docroot/sites/default/files/
```

which works. The important part is the `/` at the end of files on the remote path. That will sync down the content of files. Of course, you then have to add files to the local path. The `-l` on the original rsync command is to copy symlinks as symlinks which includes `files`.

>        --links, -l              copy symlinks as symlinks

This production environment is a Next instance.
I'm running `Acquia CLI 1.19.0` as reported by ` acli --version`

### PR DESCRIPTION
**Pull request by grasmash**
**Motivation**

Fixes #731

**Proposed changes**

Add the -z option to rsync to compress the files prior to transfer.

**Testing steps**

3. `acli pull:files`

**Merge requirements**

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -651,7 +651,14 @@ abstract class PullCommandBase extends CommandBase {
     $this->localMachineHelper->getFilesystem()->mkdir($destination);
     $command = [
       'rsync',
-      '-rltDvPhe',
+      // -a archive mode; same as -rlptgoD.
+      // -z compress file data during the transfer.
+      // -v increase verbosity.
+      // -P show progress during transfer.
+      // -k transform symlink to a dir into referent dir.
+      // -h output numbers in a human-readable format.
+      // -e specify the remote shell to use.
+      '-avPhekz',
       'ssh -o StrictHostKeyChecking=no',
       $chosen_environment->sshUrl . ':' . $source_dir,
       $destination,
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -658,7 +658,7 @@ abstract class PullCommandBase extends CommandBase {
       // -k transform symlink to a dir into referent dir.
       // -h output numbers in a human-readable format.
       // -e specify the remote shell to use.
-      '-avPhekz',
+      '-avPhkze',
       'ssh -o StrictHostKeyChecking=no',
       $chosen_environment->sshUrl . ':' . $source_dir,
       $destination,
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/Command/Pull/PullCommandBase.php [hunk 1]
```diff
@@ -651,7 +651,14 @@ protected function rsyncFilesFromCloud($chosen_environment, Closure $output_call
...
       'rsync',
-      '-rltDvPhe',
+      // -a archive mode; same as -rlptgoD.
+      // -z compress file data during the transfer.
+      // -v increase verbosity.
+      // -P show progress during transfer.
+      // -k transform symlink to a dir into referent dir.
+      // -h output numbers in a human-readable format.
+      // -e specify the remote shell to use.
+      '-avPhekz',
```
danepowell, line 661: Since the `-e` option takes an argument (the shell to use, i.e. ssh), doesn't it need to be the last option in the series?

danepowell:
I don't feel good about this solution: it could have side effects, and it doesn't address the root cause of this issue, which is that we're using the wrong path to the files directory on ACN. I think it would be better to fix `getCloudSitesPath()` to return the canonical files directory: `/mnt/data/${AH_ENVIRONMENT_UUID}-real-shared/sites/default/files`

The side effect is that as symlinks are expanded into full directories, file usage could explode. i.e. every symlink in the files directory will effectively double the size of that directory once rsync'd.

I do like the idea of consolidating the rsync options, we can keep that, but without `-k`.

danepowell:
Here's an alternative approach that fixes the sites directory on ACN: https://github.com/acquia/cli/pull/744

src/Command/Pull/PullCommandBase.php
grasmash: yeeeaaah.

grasmash:
Ok, I removed `-k`. I'd still like to make this change, particularly adding `-z` to compress files for transfer.

danepowell (approved):

PR #743, opened 2021-12-09
Fix acli new failing due to ACMS patch

danepowell:
**Motivation**
`acli new` is failing because it tries to install Drupal 9.3, but ACMS includes a patch that fails to apply to Drupal 9.3

**Proposed changes**
Stop running `composer update` so that we get the known-good dependencies provided by the project templates. As soon as ACMS (and all other packages) are compatible with Drupal 9.3, the templates will update their dependencies to match.

As a happy side effect, this should make `acli new` an order of magnitude faster.

**Testing steps**

3. Run `acli new`, verify that you get Drupal 9.2 + ACMS with the recommended template.

**Merge requirements**

### DIFF
--- a/src/Command/NewCommand.php
+++ b/src/Command/NewCommand.php
@@ -20,9 +20,9 @@ class NewCommand extends CommandBase {
   /**
    * {inheritdoc}.
    */
-  protected function configure() {
+  protected function configure(): void {
     $this->setDescription('Create a new Drupal project')
-      ->addOption('distribution', NULL, InputOption::VALUE_REQUIRED, 'The composer package name of the Drupal distribution to download')
+      ->addOption('distribution', NULL, InputOption::VALUE_REQUIRED, 'The Composer package name of the Drupal distribution to download')
       ->addArgument('directory', InputArgument::OPTIONAL, 'The destination directory');
   }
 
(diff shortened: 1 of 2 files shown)


src/Command/NewCommand.php
```diff
@@ -52,21 +52,10 @@ protected function execute(InputInterface $input, OutputInterface $output) {
       $dir = Path::makeAbsolute('drupal', getcwd());
     }
 
-    $output->writeln('<info>Creating project. This may take a few minutes</info>');
+    $output->writeln('<info>Creating project. This may take a few minutes.</info>');
     $this->localMachineHelper->checkRequiredBinariesExist(['composer']);
     $this->createProject($project, $dir);
 
-    if (strpos($project, 'acquia/drupal-minimal-project') !== FALSE) {
-      $this->requireDrush($dir);
```
danepowell, line 60: Drupal Minimal Project already provides Drush so we don't need to explicitly add it. And since we're no longer modifying dependencies, we don't need to run composer update.

anavarre (approved):
Confirmed `acli new` works fine now.

src/Command/NewCommand.php
grasmash: ok great.

PR #751, opened 2021-12-13
Fix rsync on Cloud next

danepowell:


src/Command/CommandBase.php
```diff
@@ -1432,7 +1432,7 @@ protected function getCloudSites($cloud_environment): array {
 
   protected function getCloudSitesPath($cloud_environment, $sitegroup) {
     if ($cloud_environment->platform === 'cloud-next') {
-      $path = "/mnt/data/{$cloud_environment->uuid}-real-shared/sites";
+      $path = "/home/clouduser/dev/sites";
```
grasmash, line 1435: Is it always going to be "dev" or is "dev" variable? Like, might it be `/home/clouduser/prod/sites` sometimes?

grasmash: I just checked. This is variable. It's something like `/home/clouduser/${env_name}/sites`

grasmash:
```suggestion
      $path = "/home/clouduser/{$cloud_environment->name}/sites";
```

PR #753, opened 2021-12-13
Update all Composer dependencies.

grasmash:
**Merge requirements**


danepowell:
I think these were not updated by Violinist because they aren't root-level dependencies. Ordinarily they'd be updated whenever their parent (root) dependencies are updated.

Issue #769, opened 2021-12-27
OpenAPI spec: VCS URI format is incorrect


manarth:
The spec defines the VCS uri as follows:

```yaml
        url:
          type: string
          description: The VCS url.
          format: uri
```

An example response from the API could be:

```
            "url": "myapplication@svn-12345.prod.hosting.acquia.com:myapplication.git"
```

The API responses contain an SSH-formatted Git identity, which does not match the specification format of `uri`.

grasmash:
I've created a corresponding internal ticket to track this at the canonical source.

PR #770, opened 2021-12-27
Fix 769: remove the "format" definition for the VCS URI parameter.

manarth:
**Motivation**
Fixes [#769](https://github.com/acquia/cli/issues/769)

**Proposed changes**
Removes the `format` specification for the VCS URL parameter in the OpenAPI spec.
This stops clients auto-generated from the OpenAPI specification from raising an error when environments are listed. 

**Alternatives considered**
- Change the upstream API to output URIs. This is not recommended as its a breaking change.
- Replace `format: uri` with `pattern: ` followed by a regex. See [OpenAPI pattern documentation](https://swagger.io/docs/specification/data-models/data-types/#pattern). Not implemented because it's not clear whether the backend response is expected to be consistent for all users.

NB: this may not be relevant to the `acquia/cli` repository, if the canonical source of the OpenAPI spec is managed elsewhere.

### DIFF
--- a/assets/acquia-spec.yaml
+++ b/assets/acquia-spec.yaml
@@ -3166,7 +3166,6 @@ components:
         url:
           type: string
           description: The VCS url.
-          format: uri
     DeployCodeRequest:
       type: object
       required:


grasmash:
Yes the canonical source is managed elsewhere, but we can update it there. @itafroma

grasmash:
Thank you for doing this work.

I think it would be simplest to continue documenting the discrepancies here. I can file a corresponding internal ticket that simply reference this public issue.

manarth:
Will do – it'd be good to keep our TAM @ghazlewood in the loop on the internal discussions.

manarth:
Additional change: add `1x` as a value to the `size` enum. ODE environments report their size as `1x`.

```yaml
        size:
          type: string
          description: The size of the environment. Will be null if the environment type does not support sizes.
          enum:
            - small
            - medium
            - large
            - unknown
            - 1x
```

anavarre:
FYI we have filed an internal ticket to get this fixed.

itafroma:
Hi @manarth , thanks for the report. Please do not use or rely on the spec contained inside ACLI for building outside tooling: we cannot guarantee its stability or its fitness outside of its use within ACLI and it's subject to change/break without notice. There will be a separate announcement if and when we publish the spec for public use. Sorry about the confusion!

PR #847, opened 2022-03-11
Improve generated docs.

grasmash:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**

### DIFF
--- a/src/Command/Api/ApiCommandHelper.php
+++ b/src/Command/Api/ApiCommandHelper.php
@@ -531,6 +531,7 @@ class ApiCommandHelper {
         $command->setResponses($schema['responses']);
         $command->setServers($acquia_cloud_spec['servers']);
         $command->setPath($path);
+        $command->setHelp("For more help, see https://cloudapi-docs.acquia.com/");
         $this->addApiCommandParameters($schema, $acquia_cloud_spec, $command);
         $api_commands[] = $command;
       }
--- a/src/Descriptor/ReStructuredTextDescriptor.php
+++ b/src/Descriptor/ReStructuredTextDescriptor.php
@@ -93,6 +93,8 @@ class ReStructuredTextDescriptor extends MarkdownDescriptor
     ];
     $non_default_options = [];
     foreach ($definition->getOptions() as $option) {
+      // Skip global options.
+      // @todo Show these once at the beginning.
       if (!in_array($option->getName(), $global_options)) {
         $non_default_options[] = $option;
       }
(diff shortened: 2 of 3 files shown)


anavarre:
Progress, thanks. I'm still seeing a number of deal breakers that I'm documenting below.

1. **Sidebar**

sidebar

We're seeing as many `Usage`, `Options` and `Arguments` in the sidebar as there are in code with:

```
Usage
=====
```

This is because the header level is wrong. If you try the following (`h4`), for example, it fixes the issue immediately.

```
Usage
~~~~~
```

Caveat : ideally, for SEO purposes we wouldn't have a header there, but I'm not sure we can just increase the font with RST. Might need to research this.

2. **`self` commands**

anchor-alias

- We're missing a ~header for `self` commands. For API commands you have `**api:**` and we need an equivalent.
- For whatever reason there's a rendering issue for some commands e.g. `` `command`_`` while they _all_ have an identical formatting in code.
- Some `self` commands are not returned. Their alias is. Pretty odd.

3. Missing "Command usage" header

Just above:

```
``clear-caches``
****************
```

We need a new `h2` header  to introduce that after reviewing the commands we're now getting into the details.

```
Commands usage
--------------
```

4. Display improvements and illogical order of things

Take this snippet as an example:

```
``clear-caches``
****************

Clears local Acquia CLI caches

Usage
------

- ``clear-caches``
- ``cc``
- ``cr``

Clears local Acquia CLI caches
```

* The command description is displayed twice. We should ideally get rid of the one at the bottom.
* The description needs to stand out. Ideally we'd have everything on the same line:

```
``clear-caches``: Clears local Acquia CLI caches
```

I'd even suggest we make it a `h3` header so both the command and the accompanying description immediately catch the eye.

5. Static / Dynamic installation issues

installation

I think the only way to make this block acceptable is by having a RST `.. note::` because otherwise it's pretty confusing. On top of that the header is too large and the code blocks are formatted as quotes.

Everywhere we have multiline code snippets, we should use `.. code-block:: text`

-----

Okay, let's start with that before we further improve the display.

grasmash:
1. I downsized the `usage` header to <h3>. Let me know if that works. It's not straightforward.
2. `self` commands are now listed. Aliases are now hidden.
3. Added "Commands" header after table of contents. This was actually required for the fix in #1. 
4. Command description is no longer displayed twice.

grasmash:
5. Completion command has been hidden.

PR #852, opened 2022-03-15
Adding other products as well for doc.

joshirohit100:


joshirohit100:
First stab with only few product links. Will also add other products link
Couldn't find any existing test case for this com
Screenshot 2022-03-15 at 9 44 50 PM
mand

grasmash:
Why only these 5?

joshirohit100:
I will add others as well in this PR. Just wanted to check/confirm if this is fine the operation flow of the command as expected (or any feedback) :)

src/Command/DocsCommand.php
```diff
@@ -26,7 +27,18 @@ protected function configure() {
    * @return int
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
-    $this->localMachineHelper->startBrowser('https://docs.acquia.com/acquia-cli/');
+    $acquia_products = [
+      'Acquia ACLI' => 'https://docs.acquia.com/acquia-cli/',
```
anavarre, line 31: Should be "Acquia CLI"

src/Command/DocsCommand.php
```diff
@@ -26,7 +27,18 @@ protected function configure() {
...
+      'Acquia CMS' => 'https://docs.acquia.com/acquia-cms/',
```
anavarre, line 32: We probably should have a base URL value so we can change `docs.acquia.com` in one place should we need to.

src/Command/DocsCommand.php
```diff
@@ -26,7 +27,18 @@ protected function configure() {
...
+      'Code Studio' => 'https://docs.acquia.com/code-studio/',
+      'Campaign Studio' => 'https://docs.acquia.com/campaign-studio/',
+      'Content Hub' => 'https://docs.acquia.com/contenthub/',
+    ];
+
+    $labels = array_keys($acquia_products);
+    $question = new ChoiceQuestion('Please select the Acquia Product', $labels, $labels[0]);
```
anavarre, line 39: I wonder whether we should permit doing something like `acli doc <product>` to not be prompted, similarly to what we've done with e.g. `acli ide:*`

anavarre (approved):

src/Command/DocsCommand.php
joshirohit100: Updated!

src/Command/DocsCommand.php
joshirohit100: updated!

src/Command/DocsCommand.php
joshirohit100: We can but I am not sure what value user will enter for a product. Example - for ACMS, user could type acms, acquia_cms etc. We can for sure check the value in existing list and if not there, prompt user to choose. But I feel most of the time it will be prompt (value won't match)

anavarre: Maybe we can create aliases. You mentioned `acms` or `acquia_cms`, why not have both? Maybe I'm overthinking it, but if I'm to use ACLI to open up docs, it'd better have to be blazing fast otherwise I might just go to my browser window instead.

grasmash:
Needs a test too.

joshirohit100:
Test added

src/Command/DocsCommand.php
joshirohit100: Added argument with product alias

anavarre: Lovely, thanks!

src/Command/DocsCommand.php
```diff
@@ -26,7 +31,82 @@ protected function configure() {
...
+        'url' => 'acquia-cli',
+        'alias' => ['cli', 'acli'],
+      ],
+      'Acquia CMS' => [
+        'url' => 'acquia-cms',
+        'alias' => ['acquia_cms', 'acms'],
+      ],
+      'Code Studio' => [
+        'url' => 'code-studio',
+        'alias' => ['code_studio', 'codestudio'],
```
anavarre, line 45: Also referred to as `cs`. Might be a good idea to add it too.

src/Command/DocsCommand.php
```diff
@@ -26,7 +31,82 @@ protected function configure() {
...
+      ],
+      'Campaign Studio' => [
+        'url' => 'campaign-studio',
+        'alias' => ['campaign-studio', 'campaignstudio'],
+      ],
+      'Content Hub' => [
+        'url' => 'contenthub',
+        'alias' => ['contenthub'],
```
anavarre, line 53: Also referred to as `ch`.

src/Command/DocsCommand.php
```diff
@@ -26,7 +31,82 @@ protected function configure() {
    * @return int
    */
   protected function execute(InputInterface $input, OutputInterface $output) {
-    $this->localMachineHelper->startBrowser('https://docs.acquia.com/acquia-cli/');
+    $acquia_products = [
```
anavarre, line 34: Looks like we're still missing those products:

* BLT
* Cloud Platform
* Commerce Manager
* Lando
* Acquia DAM
* Acquia DAM Classic
* Personalization
* Search
* Campaign Factory
* Campaign Studio
* Platform APIs
* Remote Administration

joshirohit100: `Campaign Studio` is already in.
`Search` is already in. 
`Commerce Manager` I didn;t add as this is because AFAIK we now don;t support/suggest commerce manager.
`Lando` Not added as lando is not acquia product + on lando page https://docs.acquia.com/lando/ we recommend to use cloud IDE

`Platform APIs` and `Remote Administration` - couldn't find link/info on https://docs.acquia.com/

PR updated for other products

src/Command/DocsCommand.php
```diff
@@ -16,7 +18,10 @@ class DocsCommand extends CommandBase {
    * {inheritdoc}.
    */
   protected function configure() {
-    $this->setDescription('Open Acquia CLI documentation in a web browser');
+    $this->setDescription('Open Acquia products documentation in a web browser')
+      ->addArgument('acquiaProduct', InputArgument::OPTIONAL, 'Name of the acquia product')
```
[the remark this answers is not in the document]
joshirohit100: updated!

src/Command/DocsCommand.php
joshirohit100: updated!

src/Command/DocsCommand.php
joshirohit100: updated!

tests/phpunit/src/Commands/DocsCommandTest.php
```diff
@@ -0,0 +1,203 @@
...
+    return $this->injectCommand(DocsCommand::class);
+  }
+
+  /**
+   * Tests the 'docs' command for ACLI.
+   *
+   * @throws \Exception
+   */
+  public function testAcliDocCommand(): void {
+    $this->docCommandValidate([0], '  [0 ] Acquia CLI');
```
grasmash, line 30: Need to switch this to use php unit data providers.

joshirohit100: PR updated with data provider for the test.

PR #859, opened 2022-03-16
CLI-703: Use defined output for processes

danepowell:
**Motivation**

Our tests always include bogus output:
```
> phpunit tests/phpunit -vvv
PHPUnit 9.5.19 #StandWithUkraine

Runtime:       PHP 8.0.16 with PCOV 1.0.11
Configuration: /Users/dane.powell/src/acquia/cli/phpunit.xml.dist

    ✔ Testing!
...  63 / 198 ( 31%)
............................................................... 126 / 198 ( 63%)
............................................................... 189 / 198 ( 95%)
..cat: https://google.com: No such file or directory
.......                                                       198 / 198 (100%)
```

The presence of `cat: https://google.com: No such file or directory` and the checkmark indicates that some of our helpers aren't using the BufferedOutput provided by tests. They also point to a larger issue, which is that these helpers directly invoke a TTY, ignoring any provided output object, which could cause all kinds of unexpected behavior (such as hanging waiting for input during tests, *cough* #858).

We should always disable TTY during tests, but that's not possible today because it would also disable command output.

**Proposed changes**

Stop abusing Process::setTty(). Decouple TTY from command output by passing a callback to the Process to stream output regardless of TTY.

**Testing steps**

3. Run `composer test`
4. Notice that `cat: https://google.com: No such file or directory` goes away
5. Run `wacli remote:drush foo.dev si` and verify you can still interact with the process (demonstrating TTY still works)

**Merge requirements**


src/Helpers/LocalMachineHelper.php
```diff
@@ -154,10 +151,16 @@ private function configureProcess(Process $process, $cwd = NULL, $print_output =
    *
    * @return \Symfony\Component\Process\Process
    */
-  protected function executeProcess(Process $process, $callback = NULL, $timeout = NULL): Process {
-    $process->setTimeout($timeout);
-    $process->start(NULL);
-    $process->wait($callback);
+  protected function executeProcess(Process $process, $callback = NULL, $print_output = TRUE): Process {
```
danepowell, line 154: `$timeout` was unused, so I removed it

grasmash:
Just make sure the ssh and Drush commands still work.

danepowell:
I tweaked it so TTY will still be used (if available) under normal conditions. TTY is now explicitly disabled only during tests.

danepowell:
I've verified that this works as expected for Drush and SSH commands. I can still run `acli remote:drush foo.dev si` and it still allows for interactive input.

src/Helpers/LocalMachineHelper.php
```diff
@@ -154,9 +152,13 @@ private function configureProcess(Process $process, $cwd = NULL, $print_output =
    *
    * @return \Symfony\Component\Process\Process
    */
-  protected function executeProcess(Process $process, $callback = NULL, $timeout = NULL): Process {
-    $process->setTimeout($timeout);
-    $process->start(NULL);
+  protected function executeProcess(Process $process, $callback = NULL, $print_output = TRUE): Process {
+    if ($callback === NULL) {
+      $callback = function ($type, $buffer) {
+        $this->output->write($buffer);
```
grasmash, line 158: Won't this print output when $print_output is false?

danepowell: Great catch! Our tests should have caught this. I added a test and fixed it to respect `$print_output`.

danepowell:
I've revised this to be 100% backwards compatible, it should really only affect tests now, since that's pretty much the only time that TTY is explicitly false but we still expect output.

PR #873, opened 2022-03-26
Allow code studio project configure/create from local machine

joshirohit100:
ACLI command to run on local to setup/configure CS project

```
acli cs:wizard --gitlab-host-name=code.acquia.com --gitlab-token=<TOKEN_HERE>
```


src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -72,6 +72,7 @@ protected function configure() {
       ->addOption('secret', NULL, InputOption::VALUE_REQUIRED, 'The Cloud Platform API secret that Code Studio will use')
       ->addOption('gitlab-token', NULL, InputOption::VALUE_REQUIRED, 'The GitLab personal access token that will be used to communicate with the GitLab instance')
       ->addOption('gitlab-project-id', NULL, InputOption::VALUE_REQUIRED, 'The project ID (an integer) of the GitLab project to configure.')
+      ->addOption('gitlab-host-name', NULL, InputOption::VALUE_REQUIRED, 'The GitLab host name.')
```
anavarre, line 75:
```suggestion
      ->addOption('gitlab-host-name', NULL, InputOption::VALUE_REQUIRED, 'The GitLab hostname.')
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -96,6 +97,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
         "Unable to authenticate with Code Studio",
         "Did you set a valid token with the <options=bold>api</> and <options=bold>write_repository</> scopes?",
         "Try running `glab auth login` to re-authenticate.",
+        "Or run command with `--gitlab-token` option. Example `acli cs:wizard --gitlab-token=TOKEN_HERE`",
```
anavarre, line 100:
```suggestion
        "Alternatively,  pass the `--gitlab-token` option. Example: `acli cs:wizard --gitlab-token=<TOKEN>`",
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -189,8 +191,11 @@ protected function commandRequiresAuthentication(InputInterface $input): bool {
    */
   protected function validateEnvironment() {
     //$this->requireCloudIdeEnvironment();
-    if (!getenv('GITLAB_HOST')) {
-      throw new AcquiaCliException('The GITLAB_HOST environmental variable must be set.');
+    // If this is cloud ide environment.
```
anavarre, line 194:
```suggestion
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -189,8 +191,11 @@ protected function commandRequiresAuthentication(InputInterface $input): bool {
...
+    if (self::isAcquiaCloudIde()) {
+      if (!getenv('GITLAB_HOST')){
+        throw new AcquiaCliException('The GITLAB_HOST environmental variable must be set or use `--gitlab-host-name` option.');
```
anavarre, line 197:
```suggestion
        throw new AcquiaCliException('The GITLAB_HOST environment variable must be set or the `--gitlab-host-name` option must be passed.');
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -189,8 +191,11 @@ protected function commandRequiresAuthentication(InputInterface $input): bool {
    */
   protected function validateEnvironment() {
     //$this->requireCloudIdeEnvironment();
-    if (!getenv('GITLAB_HOST')) {
-      throw new AcquiaCliException('The GITLAB_HOST environmental variable must be set.');
+    // If this is cloud ide environment.
+    if (self::isAcquiaCloudIde()) {
+      if (!getenv('GITLAB_HOST')){
```
anavarre, line 196:
```suggestion
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -189,8 +191,11 @@ protected function commandRequiresAuthentication(InputInterface $input): bool {
    */
   protected function validateEnvironment() {
     //$this->requireCloudIdeEnvironment();
-    if (!getenv('GITLAB_HOST')) {
-      throw new AcquiaCliException('The GITLAB_HOST environmental variable must be set.');
+    // If this is cloud ide environment.
+    if (self::isAcquiaCloudIde()) {
```
anavarre, line 195:
```suggestion
    if (self::isAcquiaCloudIde() && !getenv('GITLAB_HOST')) {
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -305,6 +310,7 @@ protected function getGitLabToken(string $gitlab_host): string {
       "* Create a token and grant it both <comment>api</comment> and <comment>write repository</comment> scopes",
       "* Copy the token to your clipboard",
       "* Run <comment>glab auth login --hostname=$gitlab_host</comment> and paste the token when prompted",
+      "* Or use <comment>--gitlab-token</comment> option with the command. Example <comment>acli cs:wizard --gitlab-token=TOKEN_HERE</comment>",
```
anavarre, line 313:
```suggestion
      "* Or pass <comment>--gitlab-token</comment> with the <comment>cs:wizard</comment> command. Example: <comment>acli cs:wizard --gitlab-token=<TOKEN></comment>",
```

anavarre:
I haven't tested the command yet but it looks like we should indicate the dependency on `glab` as well.

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated!

joshirohit100:
So on local, if we run command like - `acli cs:wizard --gitlab-host-name=code.acquia.com --gitlab-token=<TOKEN_HERE>`
we don't need glab

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -96,6 +97,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
         "Unable to authenticate with Code Studio",
         "Did you set a valid token with the <options=bold>api</> and <options=bold>write_repository</> scopes?",
         "Try running `glab auth login` to re-authenticate.",
+        "Alternatively,  pass the `--gitlab-token` option. Example: `acli cs:wizard --gitlab-token=<TOKEN>`",
```
grasmash, line 100:
```suggestion
        "Alternatively, pass the <options=bold>--gitlab-token</> option.",
```

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -189,8 +191,8 @@ protected function commandRequiresAuthentication(InputInterface $input): bool {
    */
   protected function validateEnvironment() {
     //$this->requireCloudIdeEnvironment();
-    if (!getenv('GITLAB_HOST')) {
-      throw new AcquiaCliException('The GITLAB_HOST environmental variable must be set.');
+    if (self::isAcquiaCloudIde() && !getenv('GITLAB_HOST')) {
+      throw new AcquiaCliException('The GITLAB_HOST environment variable must be set or the `--gitlab-host-name` option must be passed.');
```
grasmash, line 195: But you're not actually checking the option here? It seems like it could still fail even with --gitlab-host-name being passed.

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -305,6 +307,7 @@ protected function getGitLabToken(string $gitlab_host): string {
       "* Create a token and grant it both <comment>api</comment> and <comment>write repository</comment> scopes",
       "* Copy the token to your clipboard",
       "* Run <comment>glab auth login --hostname=$gitlab_host</comment> and paste the token when prompted",
+      "* Or pass <comment>--gitlab-token</comment> option with <comment>cs:wizard</comment> command. Example: <comment>acli cs:wizard --gitlab-token=<TOKEN></comment>",
```
grasmash, line 310: Wouldn't they also need to pass --gitlab-host-name ?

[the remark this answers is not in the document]
joshirohit100: PR updated!

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: PR updated as per suggestion.

joshirohit100: I don;t think we need that here. Reason being this error will only be triggered once user has both hostname and token.
So `validateEnvironment()` is triggered and then we check for hostname. If host name is not available in options or in local config, we throw exception here https://github.com/acquia/cli/blob/master/src/Command/CodeStudio/CodeStudioWizardCommand.php#L326

I guess we can add that info to pass the hostname at this place

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: If hostname is not available, this will fail even before reaching this method. It will fail in https://github.com/acquia/cli/blob/master/src/Command/CodeStudio/CodeStudioWizardCommand.php#L319 where we first get the hostname

src/Command/CodeStudio/CodeStudioWizardCommand.php
joshirohit100: So we check `GITLAB_HOST` env variable only when its cloud ide. If not cloud ide (assuming local), then next execution is to fetch the hostname from options or from local config and there we throw exception.

joshirohit100:
Not sure why this is failing but failure seems not related

src/Command/CodeStudio/CodeStudioWizardCommand.php
grasmash: I don't see why we should limit it in that way. We need the host url. So, I'd check `!getenv('GITLAB_HOST') && !$this->input->getOption('gitlab-host-name')`

PR #880, opened 2022-04-01
Trigger CI workflow when a release is created

grasmash:
**Motivation**

Fixes #NNN

**Proposed changes**

**Alternatives considered**

**Testing steps**

**Merge requirements**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
   schedule:
     - cron: "0 2 * * *"
+  release:
 defaults:
   run:
     # Run Git Bash on Windows. Otherwise it uses PowerShell Core and we'd need


.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
   schedule:
     - cron: "0 2 * * *"
+  release:
```
danepowell, line 7: I think using a [create event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#create) is slightly better than a `release` event since `create` ties this to a tag being created rather than a release being created. The two events aren't always synchronous, and weirdness might ensue if you do something like unpublish a release, or pre-create a tag.

danepowell: Oh, nevermind. You are right to use a `release` event, since you can only attach assets to a release, not a tag.

danepowell (approved):
Coveralls must be losing its mind

PR #915, opened 2022-04-14
CLI-556: Mutation testing

danepowell:
**Motivation**

Our test coverage is extremely high, but that's no indication of test quality, since merely running code while asserting nothing can generate 100% coverage. I happen to think our test quality is _also_ extremely high, but mutation testing is a way of ensuring that 😄 

**Proposed changes**

- Add a GitHub Action workflow to catch mutants generated by pull requests. This is the main use case, since there's likely so many existing mutants that it would be overwhelming to scan the whole codebase every time. See a demo of this: https://github.com/acquia/cli/pull/920/files
- Add a `composer infection` command that can be run locally to scan the entire codebase if you _really_ want to. This will generate something similar to a code coverage statistic, called Mutation Score Indicator (MSI). Ours is ~75%.

**Testing steps**

**Merge requirements**


composer.json
```diff
@@ -173,6 +173,12 @@
         "box-compile": [
             "php build/box.phar compile"
         ],
+        "infection-install": [
+            "curl -f -L https://github.com/infection/infection/releases/download/0.26.0/infection.phar -o build/infection.phar"
```
anavarre, line 177: Any reason not to pull the latest version always?

infection.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "https://raw.githubusercontent.com/infection/infection/0.26.0/resources/schema.json",
```
anavarre, line 2: Same question, I guess.

.github/workflows/mutation.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    name: Mutation Testing Code Review Annotations
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
```
anavarre, line 17: Can we and should we test multiple PHP versions?

anavarre (approved):
Leaving it up for Matt to comment on the actual code but https://github.com/acquia/cli/pull/920/files looks great.

composer.json
danepowell: I'd just prefer to keep the version locked so nothing breaks unexpectedly. Box in particular would break almost every time it updated.

Ultimately we should move this and Box to some other package manager like [Phive](https://phar.io/) so we can manage versions properly.

infection.json
danepowell: This is defining the schema for the Json file, you always want to keep that pinned to a static version or the schema could break.

.github/workflows/mutation.yml
danepowell: There's no benefit to testing in any particular PHP version, much less multiple versions. Mutators act on very low-level language features that aren't going to be version-dependent (if you remove a method call or flip a return value, the effect will be the same regardless of PHP version)

composer.json
grasmash, line 177: Why use a phar file rather than a composer require?

danepowell: They [don't recommend installing locally via Composer](https://infection.github.io/guide/installation.html). I'm not sure why, and I didn't feel like investigating 😄

PR #916, opened 2022-04-14
CLI-745: Parallel unit tests

### LINKED ISSUE #914: CLI-745: Run PHPUnit tests in parallel
Our test suite is getting large enough that it's mildly annoying to wait for it to finish. We could use https://github.com/paratestphp/paratest to speed it up by an order of magnitude. We'd need to fix our tests to be completely independent of each other, I think they might depend on a common fixture directory right now. Isolating tests is something we should really do regardless.

danepowell:
**Motivation**

Fixes #914

This reduces unit test suite time from 80 s to 20 s on my machine.

**Proposed changes**

- Run unit tests in parallel with paratest.
- Isolate test fixtures per-test to avoid race conditions.
- Remove extraneous code related to cache- and fixture-management now that tests run with isolated fixtures
- Forbid the use of sys_get_temp_dir, which is not thread-safe, and instead provide a safe $tmpDir via DI
- Split Coveralls reporting into a separate GitHub job since paratest behaves slightly differently with coverage enabled

**Testing steps**

3. Run `composer test` and verify that tests run in reduced time
4. Check for regressions in common ACLI commands

**Merge requirements**


phpunit.xml.dist
```diff
@@ -5,9 +5,6 @@
     <include>
       <directory suffix=".php">src</directory>
     </include>
-    <report>
```
danepowell, line 8: The presence of a `<report>` element when xdebug is disabled causes phpunit to grumble and paratest to fall over completely. Best to omit it and only enable reporting/coverage when it's needed.

danepowell:
@grasmash do you think this is worth the effort and additional complexity? The benefits are reducing unit test time (by more than 75%) and making tests more reliable by isolating each test to a unique set of fixture and tmp directories.

grasmash:
This doesn't seem like much extra complexity. Most of this code is just getting and setting the temp directory. So, probably worth it. Though, I don't ever actually have an issue with the unit tests being too slow. I typically only run tests from a single file when developing locally. Otherwise I let GitHub run the full suite.

grasmash:
Two pieces of feedback come to mind:

1. This should be submitted against the 2.X branch.
2. I would prefer not to inject an additional argument into all of the commands. It seems like getting the location of the temporary directory is a method that belongs pretty squarely in the local machine helper. Can we just use that service?

anavarre:
Can you submit against 2.x @danepowell ?

danepowell:
Yes I'll finish this today, I just ran out of time yesterday.

PR #935, opened 2022-04-26
Run coverage tests independently

danepowell:


phpunit.xml.dist
```diff
@@ -14,9 +14,6 @@
     <include>
       <directory suffix=".php">src</directory>
     </include>
-    <report>
```
danepowell, line 17: The main point of this PR is to get rid of this line, which causes PhpUnit to throw a warning and other programs (like paratest) to fatally error. It's better to run coverage independently from automated tests.

PR #936, opened 2022-04-26
CLI-747: Isolate test fixtures

danepowell:
**Motivation**

Matt and I determined that Kernel tests were reading your actual `~/.acquia/cloud_api.conf` instead of the one provided by the fixture. This was causing tests to fail locally for me.

**Proposed changes**

- Refactor kernel tests to use the test base consistently
- Refactor _all_ tests to completely isolate fixture directories between test runs (removing the need for teardown steps)

Don't let the decrease in coverage fool you, that code was _never_ covered since that was just the kernel test reading `~/.acquia/cloud_api.conf`.


tests/phpunit/src/TestBase.php
```diff
@@ -201,15 +201,15 @@ protected function setUp($output = NULL): void {
     $this->setClientProphecies();
     $this->setIo($input, $output);
 
-    $this->fixtureDir = realpath(__DIR__ . '/../../fixtures');
+    $this->fixtureDir = $this->getTempDir();
+    $this->fs->mirror(realpath(__DIR__ . '/../../fixtures'), $this->fixtureDir);
```
danepowell, line 205: This is where the magic happens, mirroring the fixture to a temporary directory rather than modifying it in-situ so there's no chance of cross-talk between tests

anavarre:
Can we not decrease test coverage?

danepowell:
The test coverage percentage is decreasing, but the actual coverage is not. The fact that the code was executed in tests (and thus "covered") was entirely a fluke.

Anyway, I'll try to add a new test case or two to make up for it.

src/CloudApi/CloudCredentials.php
```diff
@@ -58,12 +58,6 @@ public function getCloudKey(): ?string {
       return $this->datastoreCloud->get('acli_key');
     }
 
-    // Legacy format.
```
danepowell, line 61: @grasmash legacy cloud configs (with a "key" instead of "keys") throw an error. Did you intentionally deprecate that format?

If so, I'll remove all associated legacy code like this.

If not, please restore support and I'll add a test case (as below)

PR #947, opened 2022-04-29
Kill escaped mutants

danepowell:
Just trying my hand at this... finding mutants turns out to be a lot easier than killing them 😄 

To verify this is helping:

- Observe escaped mutants like this one in prior test runs: https://github.com/acquia/cli/runs/6231672360?check_suite_focus=true#step:6:85
- Observe that this mutant no longer exists on this PR: https://github.com/acquia/cli/runs/6233376232?check_suite_focus=true#step:6:34

You can also run mutation tests locally and compare before and after: `composer infection-install && ./build/infection.phar --filter=AcsfClientService.php -j2 --only-covered`

### DIFF
--- a/src/AcsfApi/AcsfClientService.php
+++ b/src/AcsfApi/AcsfClientService.php
@@ -42,6 +42,7 @@ class AcsfClientService extends ClientService {
 
     if (getenv('ACSF_USERNAME') && getenv('ACSF_PASSWORD') ) {
       $this->machineIsAuthenticated = TRUE;
+
       return $this->machineIsAuthenticated;
     }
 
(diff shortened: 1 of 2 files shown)


grasmash:
Can we get some overall stats on mutations ?

danepowell:
Our MSI (indicator of test quality) was about 76% on master before we merged 2.x into master. Now it's down to 36% but I think that's inaccurate because local tests are broken yet again.

You can run `composer infection-install && composer infection` at any time locally to get a full report.

grasmash:
Can we get that to be reported in each pull request the same way that we do for code coverage?

danepowell:
I'll see what I can do: #952

danepowell:
As I suspected, it takes way too long to run the full mutation test suite on PRs (20+ minutes). We can at least run it on `master` though.

PR #948, opened 2022-04-29
Use pcov instead of xdebug for coverage

danepowell:
It's twice as fast! Compare 2 m ([last run with Xdebug](https://github.com/acquia/cli/runs/6233376495?check_suite_focus=true)) to 1 m ([this run with pcov](https://github.com/danepowell/cli/runs/6235229772?check_suite_focus=true))

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -16,7 +16,6 @@ jobs:
         with:
           php-version: 8.0
           tools: composer:v2
-          extensions: pcov
           coverage: pcov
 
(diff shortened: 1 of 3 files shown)


.github/workflows/mutation.yml
```diff
@@ -16,7 +16,6 @@ jobs:
         with:
           php-version: 8.0
           tools: composer:v2
-          extensions: pcov
```
danepowell, line 19: According to the shivammathur/setup-php@v2 docs the pcov extension is installed by default

PR #963, opened 2022-05-04
Remove duplicate release drafter trigger

danepowell:
`pull_request` and `pull_request_target` triggers are essentially identical and using both in the same workflow results in multiple workflows running at once.

The only difference is that `pull_request_target` runs in the context of the base branch and injects repo secrets. Using `pull_request` is safer and just requires tweaking permissions.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -4,16 +4,14 @@ on:
   push:
     branches:
       - master
-
   pull_request:
-  pull_request_target:
 
 jobs:
   update_release_draft:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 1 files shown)


grasmash:
I agree this looks like the right configuration. However, I observed (perhaps due to a bug) that no jobs would fire if I didn't add both triggers. Maybe it'll just work now?

danepowell:
Most likely due to missing permissions, you can see that failing for the pull_request jobs today: https://github.com/acquia/cli/runs/6295369638?check_suite_focus=true

That's why I change the permissions in this PR.

Though that results in an error, not "no jobs would fire"... not sure about that.

PR #975, opened 2022-05-09
CLI-754: Fix slow PHPUnit tests

danepowell:
**Motivation**

Tests are slower than necessary. This reduces the duration of unit tests from 90 seconds to 20 seconds!

**Proposed changes**

- Stop clearing the kernel cache between runs
- Run the event loop once immediately when it's created before starting the periodic loop. This feels a little hacky but I don't know how else to fix it.

**Merge requirements**

### DIFF
--- a/tests/phpunit/src/ApplicationTestBase.php
+++ b/tests/phpunit/src/ApplicationTestBase.php
@@ -25,9 +25,6 @@ class ApplicationTestBase extends TestBase {
 
   public function setUp($output = NULL):void {
     parent::setUp($output);
-    // If kernel is cached from a previous run, it doesn't get detected in code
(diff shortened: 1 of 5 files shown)


tests/phpunit/src/ApplicationTestBase.php
```diff
@@ -25,9 +25,6 @@ class ApplicationTestBase extends TestBase {
 
   public function setUp($output = NULL):void {
     parent::setUp($output);
-    // If kernel is cached from a previous run, it doesn't get detected in code
```
danepowell, line 28: Not sure why but this seems no longer relevant. You can see our coverage stays the same in this PR.

PR #976, opened 2022-05-09
CLI-758: PHP 8.1 warnings

danepowell:
We still have a lot of PHP 8.1 compatibility issues. These mostly manifest as deprecation warnings. You can see them if you have error reporting enabled: https://github.com/danepowell/cli/runs/6359193752?check_suite_focus=true


tests/phpunit/src/Commands/Pull/PullCommandTestBase.php
```diff
@@ -15,6 +16,8 @@
  */
 abstract class PullCommandTestBase extends CommandTestBase {
 
+  use IdeRequiredTestTrait;
```
danepowell, line 19: Maybe this isn't best. If it's only used for a few test cases, it should be a helper not a trait.

danepowell:
> Why aren't these warning causing test failures?

Great question, that's why item #1 on the todos for this issue is `Automated tests should fail on warnings` 😄

.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: shivammathur/setup-php@v2
         with:
+          ini-file: development
```
danepowell, line 35: This enables reporting of deprecation errors. In general the development ini is stricter than the production ini, so I think it's good/safe to use in CI.

phpunit.xml.dist
```diff
@@ -9,6 +9,7 @@
         convertErrorsToExceptions="true"
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
+        convertDeprecationsToExceptions="true"
```
danepowell, line 12: I verified locally that this causes PHPUnit to catch the deprecation warnings

tests/phpunit/src/Commands/Ide/IdeHelper.php
```diff
@@ -0,0 +1,30 @@
+<?php
```
danepowell, line 1: The refactoring of IdeRequiredTestTrait is necessary because trait methods cannot be referenced outside of a class in PHP 8.1.

grasmash:
Need to resolve conflicts on this

PR #986, opened 2022-05-12
Fix local test failures

danepowell:


danepowell:
Nevermind, I had some cruft in my fixture directory (a leftover `.git` directory that wasn't showing up in the git status). Removing that fixed tests.

PR #1006, opened 2022-05-25
CLI-679: Consolidate internal commands under self

danepowell:
This moves commands to the following new locations:
- `app:new`
- `app:link`
- `app:unlink`
- `self:list`
- `self:telemetry:toggle`
- `self:telemetry:disable`
- `self:telemetry:enable`

It also adds aliases to preserve the original command.

Annoyingly, Symfony has marked the DumpCompletion command (`completion`) as final so we can't extend it to change the name: https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Console/Command/DumpCompletionCommand.php

We could open a PR with Symfony to change this but I'm not sure it's worth the effort.

### DIFF
--- a/src/Command/App/NewCommand.php
+++ b/src/Command/App/NewCommand.php
@@ -1,7 +1,8 @@
 <?php
 
-namespace Acquia\Cli\Command;
+namespace Acquia\Cli\Command\App;
 
+use Acquia\Cli\Command\CommandBase;
 use Acquia\Cli\Exception\AcquiaCliException;
 use Acquia\DrupalEnvironmentDetector\AcquiaDrupalEnvironmentDetector;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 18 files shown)


src/Command/App/NewCommand.php
```diff
@@ -15,15 +16,16 @@
  */
 class NewCommand extends CommandBase {
 
-  protected static $defaultName = 'new';
+  protected static $defaultName = 'app:new';
```
grasmash, line 19: This command doesn't create a new cloud app, which is what I'd expect a command named this to do.

anavarre: Aldo this will break the Cloud IDE Get Started page from which we have a CTA to create a new app.

danepowell: It shouldn't break anything since I added an alias for backwards compatibility.

So what do you want to do? Just undo the change (back to `acli new`), or rename it something like `acli app:new:local`? I would argue that `acli new` is even more confusing and ambiguous than `acli app:new`, we should add _some_ sort of context.

grasmash: I think `app:new:local` makes sense. Keep alias for `new`.

PR #1046, opened 2022-06-08
ACLI migrate command fix.

NetraPK:
This PR consist following fixes:

1. Added needs for build job, so that new "build drupal" job will be dependent on "'Build Code','Manage Secrets'" job.
2. At the end of the command it will remove the acquia-pipelines.yml file from the root repository.
3. "validateEnvironment" method was not getting called which is used to check GITLAB_HOST environment variable is set or not.so added it in "authenticateWithGitLab" method.


grasmash:
Please fill out the title and description of this PR. Convert from draft when ready for review.

PR #1057, opened 2022-06-13
Fixed bug GL-701, GL-702

NetraPK:
Below PR consist of bug fixes:
1. Changed the prompt msg for BLT questions.
2. If the job name is launch_ode it was getting migrated, Added fix to migrate other script except than launch_ode.


grasmash:
Please change test so that it asserts this code works correctly.

grasmash:
There’s a difference between executing a line of code and asserting that it’s doing what is expected. 

This pull request introduces a functional difference. There was a bug before, now there isn’t. The test needs to assert that the bug is no longer present.

So for instance, you could add a “launch_ode” script to the test fixture and then assert that the command actually removes it as expected.

PR #1068, opened 2022-06-21
GL-703 | Bug fix for acli migrate command.

NetraPK:
As discussed on GL-635 ticket, removing creation of code studio project from acli migrate command and throwing error messages.

### DIFF
--- a/src/Command/CodeStudio/CodeStudioCommandTrait.php
+++ b/src/Command/CodeStudio/CodeStudioCommandTrait.php
@@ -177,7 +177,7 @@ trait CodeStudioCommandTrait {
    *
    * @return array
    */
-  protected function determineGitLabProject(ApplicationResponse $cloud_application): array {
+  protected function determineGitLabProject(ApplicationResponse $cloud_application, Bool $create_new_cs_project = TRUE): array {
     // Use command option.
     if ($this->input->getOption('gitlab-project-id')) {
       $id = $this->input->getOption('gitlab-project-id');
@@ -201,11 +201,16 @@ trait CodeStudioCommandTrait {
     }
     // Prompt to create project.
     else {
-      $this->io->writeln([
+      $error_message = [
         "",
-        "Could not find any existing Code Studio project for Acquia Cloud Platform application <comment>{$cloud_application->name}</comment>.",
-        "Searched for UUID <comment>{$cloud_application->uuid}</comment> in project descriptions.",
-      ]);
+        "Could not find any existing Code Studio project for Acquia Cloud Platform application {$cloud_application->name}.",
+        "Searched for UUID {$cloud_application->uuid} in project descriptions.",
+      ];
+      if (!$create_new_cs_project) {
+        $this->io->error($error_message);
+        throw new AcquiaCliException("Contact an account manager for further assistance");
+      }
+      $this->io->writeln($error_message);
       $create_project = $this->io->confirm('Would you like to create a new Code Studio project? If you select "no" you may choose from a full list of existing projects.');
       if ($create_project) {
         return $this->createGitLabProject($cloud_application);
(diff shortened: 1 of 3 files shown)


src/Command/CodeStudio/CodeStudioCommandTrait.php
```diff
@@ -201,11 +201,16 @@ protected function determineGitLabProject(ApplicationResponse $cloud_application
...
         "",
-        "Could not find any existing Code Studio project for Acquia Cloud Platform application <comment>{$cloud_application->name}</comment>.",
-        "Searched for UUID <comment>{$cloud_application->uuid}</comment> in project descriptions.",
-      ]);
+        "Could not find any existing Code Studio project for Acquia Cloud Platform application {$cloud_application->name}.",
+        "Searched for UUID {$cloud_application->uuid} in project descriptions.",
+      ];
+      if (!$create_new_cs_project) {
+        $this->io->error($error_message);
+        throw new AcquiaCliException("Contact an account manager for further assistance");
```
grasmash, line 211: Is this covered in a test?

grasmash (changes requested):
This requires a test or assertion.

src/Command/CodeStudio/CodeStudioCommandTrait.php
```diff
@@ -207,11 +207,16 @@ protected function determineGitLabProject(ApplicationResponse $cloud_application
...
+      $error_message = [
         "",
-        "Could not find any existing Code Studio project for Acquia Cloud Platform application <comment>{$cloud_application->name}</comment>.",
-        "Searched for UUID <comment>{$cloud_application->uuid}</comment> in project descriptions.",
-      ]);
+        "Could not find any existing Code Studio project for Acquia Cloud Platform application {$cloud_application->name}.",
+        "Searched for UUID {$cloud_application->uuid} in project descriptions.",
+      ];
+      if (!$create_new_cs_project) {
+        $this->io->error($error_message);
```
danepowell, line 216: Please add a test for the presence of this error message so that the escaped mutant is squashed.

src/Command/CodeStudio/CodeStudioCommandTrait.php
```diff
@@ -207,11 +207,16 @@ protected function determineGitLabProject(ApplicationResponse $cloud_application
...
+        throw new AcquiaCliException("Contact an account manager for further assistance");
```
danepowell, line 217: Please change this exception to be meaningful out of context. i.e. "Could not find any existing Code Studio project, and the option to create a new project was not selected".

In fact you probably want to just put the `$error_message` here.

It's good to also say `Contact an account manager for further assistance` but that should not be part of the exception.

Make sure to also use the string substitutions in AcquiaCliException so that PII (the Cloud application name and UUID) are excluded.

tests/phpunit/src/Commands/CodeStudio/CodeStudioPipelinesMigrateCommandTest.php
```diff
@@ -133,4 +134,62 @@ public function testCommand($mocked_gitlab_projects, $inputs, $args) {
     $this->assertIsArray($contents['setup']['needs']);
   }
 
+  /**
+   * @return array
+   */
+  public function providertestdetermineGitLabProject(): array {
```
danepowell, line 140: Why use a data provider if there's only one data set?

I think you want two data sets here: one with `$create_new_cs_project = TRUE` and one with `$create_new_cs_project = FALSE`.

src/Command/CodeStudio/CodeStudioCommandTrait.php
```diff
@@ -183,7 +183,7 @@ protected function authenticateWithGitLab(): void {
    *
    * @return array
    */
-  protected function determineGitLabProject(ApplicationResponse $cloud_application): array {
+  protected function determineGitLabProject(ApplicationResponse $cloud_application, Bool $create_new_cs_project = TRUE): array {
```
danepowell, line 186: The only other code that calls this method is `codestudio:wizard` right? Please change that to pass `TRUE` explicitly and remove the default parameter value here.

tests/phpunit/src/Commands/CodeStudio/CodeStudioPipelinesMigrateCommandTest.php
danepowell: Actually since create_new_cs_project can only be true for the `codestudio:wizard` command, I think you need to add a separate test to that command, or else add an argument to the Pipelines Migrate command.

danepowell (changes requested):

danepowell:
No response in 7 days, closing as stale. Feel free to reopen when the above feedback is addressed.

PR #1070, opened 2022-06-21
DO NOT MERGE: Auto-merge dependabot and improve Composer validation

danepowell:
- Tweak our automerge action to be a little more generic, I want to be able to re-use the same script in all of our repos
- Auto-merge Dependabot PRs, which we currently get as a backup to Violinist for critical security vulns
- Improve Composer validation of PSR violations. Prevents issues like the phantom tests I caught in https://github.com/acquia/cli/pull/1065

We should wait to merge until a [Composer 2.4 release](https://github.com/composer/composer/releases) supports `--strict-psr` and remove the self-update step.


danepowell:
No point keeping this open... will reopen when we can

PR #1075, opened 2022-06-27
IDE-2333: Upgrade to ACLI 2 in IDE

danepowell:
**Motivation**

Before running external commands (namely Composer as part of `pull:code`), ACLI tries to verify that the local PHP version matches the application version. ACLI assumes that the invoked PHP version is the default PHP version for an environment. That's not always true, and causes problems if you try to pull code while invoking ACLI with a specific PHP version (e.g. `/usr/bin/php8.1 acli`).

**Proposed changes**

PHP_VERSION is the actual default PHP version in IDE environments. Use PHP_VERSION if available, otherwise fall back to the version invoking ACLI.

**Alternatives considered**

There's still some edge cases this won't catch, such as if a customer has aliased `composer` to use a specific PHP version, but it's the best we can reasonably do.

**Testing steps**

3. Run `acli pull:code` and pick an application with a different PHP version than you run locally. See that you get a warning about mismatched PHP version.
4. Run `PHP_VERSION=[x.y] wacli pull:code`, where x.y matches your remote PHP version, to simulate pulling code in an IDE with PHP_VERSION set. See that the warning is avoided.

**Merge requirements**

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -899,7 +899,7 @@ abstract class PullCommandBase extends CommandBase {
    * @return string
    */
   protected function getCurrentPhpVersion(): string {
-    return PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;
+    return getenv('PHP_VERSION') ?: PHP_MAJOR_VERSION . '.' . PHP_MINOR_VERSION;
   }
 
   /**


src/Command/Pull/PullCommandBase.php
```diff
@@ -896,10 +896,14 @@ protected function environmentPhpVersionMatches($environment): bool {
   }
 
   /**
+   * Get current PHP version using IDE-specific variables when possible.
+   *
+   * Disable mutation testing because we cannot mock PHP constants.
```
anavarre, line 901: Is it waiting on https://github.com/infection/infection/issues/315 ? If yes could you link to it?

anavarre: I just shot in the dark and that wasn't it. Ignore 😉

anavarre:
When specifically passing the `PHP_VERSION` flag, then it works.

```
test:~/project (pipelines-build-master) $ PHP_VERSION=7.4 acli pull:code
Using Cloud Application myapp

 Choose a Cloud Platform environment [Dev, dev (vcs: pipelines-build-master)]:
  [0] Dev, dev (vcs: pipelines-build-master)
  [1] Prod, prod (vcs: tags/2020-09-28)
  [2] Stage, test (vcs: pipelines-build-master)
 > 0

    ✔ Pulling code from the Cloud Platform
    ✔ Installing Composer dependencies
    ✔ Clearing Drupal caches via Drush
```

PR #1078, opened 2022-06-30
IDE-2333: Detect PHP version in IDE via config file

danepowell:
**Motivation**

#1075 did not work as intended because the `PHP_VERSION` env var is not set when invoking ACLI via a specific version of PHP (rather than via the PHP wrapper)

**Proposed changes**

Revert functional changes in #1075 and instead read default PHP version from the version-selection config file.

**Alternatives considered**

Run `php --version` to discover the actual default PHP version, but then we're left with the challenge of parsing the version string which requires some ugly bash-fu.

**Testing steps**

1. Start an IDE (needs to have an active SSH agent, so probably in staging instead of local)
2. Apply changes to acli.sh from the IDE counterpart to this PR (to invoke ACLI via PHP 8)
3. Download the PR build of ACLI from transfer.sh and modify acli.sh to use it instead of the global ACLI version
4. Ensure that your IDE runs PHP 7.4 by default, and find a Cloud application that also runs PHP 7.4
5. Run `acli pull:code` and choose the PHP 7.4 application.

You should see that ACLI runs fine and without throwing any warnings. This implies that ACLI is running via PHP 8 while still properly detecting that the environment is PHP 7.4. 

**Merge requirements**

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -10,13 +10,12 @@
   <arg name="cache" value="build/.phpcs-cache"/>
   <arg name="parallel" value="10"/>
 
-  <file>.</file>
+  <file>src</file>
+  <file>tests</file>
 
   <!-- Danger! Exclude patterns apply to the full file path, including parent directories of the current repository. -->
   <!-- Don't exclude common directory names like `build`, which will fail on Travis CI because of /home/travis/build/acquia/<project>. -->
   <!-- @see https://github.com/squizlabs/PHP_CodeSniffer/issues/981 -->
-  <exclude-pattern>var/</exclude-pattern>
-  <exclude-pattern>vendor/*</exclude-pattern>
   <exclude-pattern>tests/fixtures/*</exclude-pattern>
 
   <rule ref="AcquiaPHP" />
--- a/src/Command/Ide/IdeCommandBase.php
+++ b/src/Command/Ide/IdeCommandBase.php
@@ -4,6 +4,7 @@ namespace Acquia\Cli\Command\Ide;
 
 use Acquia\Cli\Command\CommandBase;
 use Acquia\Cli\Exception\AcquiaCliException;
+use Acquia\Cli\Helpers\IdeCommandTrait;
 use AcquiaCloudApi\Endpoints\Ides;
 use AcquiaCloudApi\Response\IdeResponse;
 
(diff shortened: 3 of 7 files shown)


.editorconfig
```diff
@@ -0,0 +1,10 @@
+# editorconfig.org
```
danepowell, line 1: Sorry for the unrelated change here, but I had to re-initialize PhpStorm and realized a bunch of these settings need to be imported

phpcs.xml.dist
```diff
@@ -10,13 +10,12 @@
...
   <arg name="parallel" value="10"/>
 
-  <file>.</file>
+  <file>src</file>
+  <file>tests</file>
 
   <!-- Danger! Exclude patterns apply to the full file path, including parent directories of the current repository. -->
   <!-- Don't exclude common directory names like `build`, which will fail on Travis CI because of /home/travis/build/acquia/<project>. -->
   <!-- @see https://github.com/squizlabs/PHP_CodeSniffer/issues/981 -->
-  <exclude-pattern>var/</exclude-pattern>
```
danepowell, line 18: I discovered a PhpStorm bug that causes it to silently fail to run PHPCS if you apply this particular exclude pattern. It's truly bizarre.

Changing the configuration to only scan source and test directories allows us to work around the bug and is more performant to boot.

src/Command/Ide/IdeCommandBase.php
```diff
@@ -12,10 +13,7 @@
  */
 abstract class IdeCommandBase extends CommandBase {
 
-  /**
-   * @var string
-   */
-  private $phpVersionFilePath;
+  use IdeCommandTrait;
```
danepowell, line 16: Need to move these methods to a trait in order to reuse them in the Pull command. Generally, I'm on a mission to move a lot of the code from command "base" classes into traits so they can be reused.

src/Command/Pull/PullCommandBase.php
```diff
@@ -861,7 +864,7 @@ protected function determineEnvironment(InputInterface $input, OutputInterface $
    */
   protected function checkEnvironmentPhpVersions(EnvironmentResponse $environment): void {
     if (!$this->environmentPhpVersionMatches($environment)) {
-      $this->io->warning("You are using PHP version {$this->getCurrentPhpVersion()} but the upstream environment {$environment->label} is using PHP version {$environment->configuration->php->version}");
+      $this->io->warning("You are using PHP version {$this->getIdePhpVersion()} but the upstream environment {$environment->label} is using PHP version {$environment->configuration->php->version}");
```
danepowell, line 867: Here's the meat of this PR, calling `getIdePhpVersion` which reads the PHP version from a config file. This should be safe because pull:code can only be called from within an IDE, and this file should always exist.

danepowell:
I'm going to merge this because it can't be more broken than #1075 and I'd rather not ship that in the next release.

Please test in conjunction with IDE-2333 and we'll follow up if necessary.

PR #1079, opened 2022-07-06
CLI-549: Allow for realm specificity

danepowell:
**Motivation**

We consider an application alias to be equivalent to an application name (nee "sitegroup"). The problem is that multiple realms can have identically-named applications, leading to undefined behavior for applications that exist across multiple realms. A fully-qualified application alias should include both the realm and name (together, the "hosting id").

**Proposed changes**

Make the application equivalent to a "hosting id", which is a combination of application realm and name in the form `realm:name`. If realm is omitted but the alias is still unique (i.e. matching only one app), the alias will work as before and return that app. If the realm is omitted and multiple applications match, throw an exception.

Theoretically it is possible for multiple applications in the _same_ realm to have an identical name, but in practice I don't think this ever happens. If it ever does happen, still throw an exception and the user will need to use an application uuid instead of an alias.

**Testing steps**

Run `acli ckc` and `acli self:clear-caches`.

```
$ acli ide:list pipelinesvalidation2
# Should return IDEs for this app without prompts or errors

$ acli ide:list pipelinesvalida
# Should throw an error due to invalid alias as well as help text about aliases

$ acli ide:list b[REDACTED]t
# Should throw an error due to a non-unique alias and list matching valid aliases as well as help text

$ acli ide:list prod:b[REDACTED]t
# Should list IDEs for this application, if any
```

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -407,9 +407,10 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
    * Add argument and usage examples for applicationUuid.
    */
   protected function acceptApplicationUuid() {
-    $this->addArgument('applicationUuid', InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias')
+    $this->addArgument('applicationUuid', InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias (i.e. a sitegroup optionally prefixed with the realm)')
       ->addUsage(self::getDefaultName() . ' [<applicationAlias>]')
       ->addUsage(self::getDefaultName() . ' myapp')
+      ->addUsage(self::getDefaultName() . ' prod:myapp')
       ->addUsage(self::getDefaultName() . ' abcd1234-1111-2222-3333-0e02b2c3d470');
 
     return $this;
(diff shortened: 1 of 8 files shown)


src/Command/CommandBase.php
```diff
@@ -1127,36 +1128,36 @@ public static function getAliasCache() {
...
+      $aliases = array_map($callback, (array) $customer_applications);
+      $this->io->error(sprintf("Multiple applications match this alias. Use one of the following aliases instead: %s", implode(', ', $aliases)));
+      throw new AcquiaCliException(sprintf("An application matching the alias %s was found in multiple hosting realms.", $application_alias));
     }
 
-    $this->logger->debug("Found application {$customer_application->uuid} matching alias $application_alias.");
+    $customer_application = $customer_applications[0];
 
-    // Remove the host=@*.$drush_site query as it would persist for future requests.
-    $acquia_cloud_client->clearQuery();
```
danepowell, line 1159: Not true, the client is generated uniquely by a factory on each call to `$this->cloudApiClientService->getClient();`

src/Command/CommandBase.php
```diff
@@ -407,9 +407,10 @@ public function run(InputInterface $input, OutputInterface $output): int {
    * Add argument and usage examples for applicationUuid.
    */
   protected function acceptApplicationUuid() {
-    $this->addArgument('applicationUuid', InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias')
+    $this->addArgument('applicationUuid', InputArgument::OPTIONAL, 'The Cloud Platform application UUID or alias (i.e. a sitegroup optionally prefixed with the realm)')
```
eporama, line 410: we should not use the term "sitegroup". That is an internal only machine name that we have been trying to move away from for a few years.

eporama: > (i.e. an application name optionally prefixed by the realm.)

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -195,7 +194,7 @@ public function testConvertInvalidApplicationAliasToUuidArgument(): void {
       $this->executeCommand(['applicationUuid' => $alias], []);
     }
     catch (AcquiaCliException $exception) {
-      $this->assertEquals("The {applicationUuid} argument must be a valid UUID or application alias that is accessible to your Cloud user.", $exception->getMessage());
+      $this->assertEquals("The {applicationUuid} argument must be a valid UUID or unique application alias accessible to your Cloud user.", $exception->getMessage());
```
anavarre, line 197: Nit: Cloud Platform user.

danepowell:
Mark Trapp says there is no human-readable application identifier that's guaranteed to be unique.

I've never seen two identically-named applications in the same realm except in the API spec, but it's theoretically possible. If a customer is unfortunate enough to be one of those applications, they'll need to always use the application uuid as an alias.

anavarre (approved):

PR #1117, opened 2022-08-01
CLI-832 | Acli tool export json file with extract characters on the command acli email:configure

klaw772:
**Motivation**

Fixes a bug where customers, when exporting JSON data, saw escaped forward slashes in DNS record values, causing these to not verify correctly when sent to a DNS provider.

**Proposed changes**

Adds the JSON_UNESCAPED_SLASHES constant to the json_encode() method used to dump a JSON representation of DNS record data.

**Testing steps**

1. Follow the [contribution guide](https://github.com/acquia/cli/blob/HEAD/CONTRIBUTING.md#building-and-testing) to set up your development environment.
3. Run `./bin/acli email:configure` and verify that when exporting a JSON file, forward slashes are no longer escaped.


danepowell (changes requested):
Thanks! Can we get an automated test to prevent this from happening again?

danepowell:
To elaborate, it looks like the createDnsText method is covered (yay!) but has zero assertions (boo!) leading to a bunch of escaped mutants:

- https://coveralls.io/builds/51325470/source?filename=src%2FCommand%2FEmail%2FConfigurePlatformEmailCommand.php#L324
- https://dashboard.stryker-mutator.io/reports/github.com/acquia/cli/main#mutant/Command/Email/ConfigurePlatformEmailCommand.php

So we _really_ need a test with a few assertions for this.

In Slack I suggested mocking the filesystem:
```
$mock_file_system = $this->mockGetFilesystem($local_machine_helper);
$mock_file_system->dumpFile($'dns-records.json', 'whatever-the-json-should-look-like')->shouldBeCalled();
```

tests/phpunit/src/Commands/Email/ConfigurePlatformEmailCommandTest.php
```diff
@@ -581,4 +686,16 @@ public function testConfigurePlatformEmailWithAlreadyEnabledEnvs($base_domain, $
 
   }
 
+  /**
+   * @param \Prophecy\Prophecy\ObjectProphecy $local_machine_helper
+   *
+   * @return \Prophecy\Prophecy\ObjectProphecy
+   */
+  protected function mockGetFilesystem(ObjectProphecy $local_machine_helper) {
```
klaw772, line 694: @danepowell I added this because the TestBase implementation of this method returns an actual FileSystem and not a Prophecy

danepowell (approved):

PR #1119, opened 2022-08-03
RADX-676 |  D7 docroot update command

rahulgupta-acquia:
**Motivation**

RADX-628

**Proposed changes**

add a new command for update d7 docroot with latest available updates.

**Alternatives considered**

**Testing steps**

1. Prepare a d7 docroot(with available updates)  in separate folder.
2. run the below command `./bin/acli app:update-d7-packages`

4. Clear the kernel cache to pick up new and changed commands: `./bin/acli ckc`
5. (add specific steps for this pr)

**Merge requirements**


src/Command/DrupalUpdate/DrupalOrgClient.php
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * @var FileSystemUtility
+   */
+  private FileSystemUtility $fileSystemUtility;
+
+  /**
+   * @param FileSystemUtility $fileSystemUtility
+   */
+  public function setFileSystemUtility(FileSystemUtility $fileSystemUtility): void {
```
danepowell, line 21:
```suggestion
  protected function setFileSystemUtility(FileSystemUtility $fileSystemUtility): void {
```

src/Command/DrupalUpdate/DrupalOrgClient.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return [];
+  }
+
+  /**
+   * @param string $project
+   * @return mixed
+   * @throws AcquiaCliException|GuzzleException
+   */
+  protected function fetchAvailablePackageReleases(string $project): mixed {
```
danepowell, line 73:
```suggestion
  private function fetchAvailablePackageReleases(string $project): mixed {
```

danepowell (changes requested):
Please kill as many mutants as you can (I'm talking about the annotations that appear on the "files changed" tab). This usually requires adding more test cases and/or adding assertions to your existing test cases. If this isn't clear, please schedule a call with me next week and I'll walk you through it.

Please also remove the bootstrap.inc file in the test fixtures.

tests/phpunit/src/Commands/DrupalUpdate/DrupalPackageManager.php
```diff
@@ -0,0 +1,316 @@
...
+    $this->io = new SymfonyStyle($input, $output);
+    $this->setFileSystemUtility(new FileSystemUtility($input, $output));
+    $this->setPackageUpdater(new PackageUpdater($input, $output));
+  }
+
+  /**
+   * @param string $drupal_project_cwd_path
+   * @return mixed
+   */
+  protected function fetchCorePackageVersion(string $drupal_project_cwd_path): mixed {
```
danepowell, line 106: Why do files in the tests directory seem to duplicate files in the src directory? For instance this method looks the same as its counterpart in the src directory.

danepowell:
Per our call, Rahul will add test case / assertions to fix the mutations related to Drupal core updates and mock the drupal.org API calls in tests.

Also TIL you cannot add labels or reviewers to a PR if you don't have push access 🤷

danepowell:
Didn't we agree to move this into a standalone CLI? Or did we agree to hide the command? Now I can't recall. But the more I think about it, this just doesn't make any sense to put into Acquia CLI. Acquia CLI is 100% consumer-focused, but this command isn't supposed to ever be used by consumers. It also requires bootstrapping Drupal, which violates a core tenet of Acquia CLI not to be coupled to Drupal installs.

shriacquia:
> Didn't we agree to move this into a standalone CLI? Or did we agree to hide the command? Now I can't recall. But the more I think about it, this just doesn't make any sense to put into Acquia CLI. Acquia CLI is 100% consumer-focused, but this command isn't supposed to ever be used by consumers. It also requires bootstrapping Drupal, which violates a core tenet of Acquia CLI not to be coupled to Drupal installs.

We agreed to put it in the ACLI but hidden by default and move it out into another repo in Q4. anticipating that a heavy PDLC process to do so would delay the functionality to be delivered. However, I now have an agreement that we need not file a new FR, instead update the existing FR with some implementation details if any. Which means that we can move to the new-repo approach upfront.

Is this PR reviewed and Ok'd technically?

danepowell:
Sorry I didn't see that Rahul had addressed my previous feedback. Yes, this is technically good to go. Great work @rahulgupta-acquia .

It sounds like this can go into a new private CLI client. If you need this merged for testing purposes in the meantime let me know, but I'd prefer to keep it out of ACLI if at all possible.

rahulgupta-acquia:
@danepowell Thanks for your review.

PR #1159, opened 2022-09-02
CLI-842: Restore single-db imports for IDE

danepowell:
@anavarre I haven't had time to give this much thought or test it, but it seems to me that this should be coupled to the IDE multisite work. If the customer has enabled multisite, ACLI should use multiple DBs, otherwise it should use a single DB.

This doesn't fix the permissions issue but it at least stops the bleeding. Let me know if you think it's an acceptable hotfix.


anavarre (approved):
I managed to get my DB to be correctly imported into `default`, thus restoring the default behavior for now.

PR #1165, opened 2022-09-12
CLI-837: Broken SSL certs cause pull:db to fail

danepowell:
**Motivation**

Fixes CLI-837

**Proposed changes**

Use a plain Guzzle Client request since we're not actually calling the Cloud API the second time but the customer domain directly. This might break some minor features that were dependent on the Cloud client, such as progress reporting, but this seems like a small price to pay to support SSL fallback. Ultimately this is all a problem with Cloud API redirecting download requests instead of streaming them directly (which is what caused the SSL issue in the first place).

Shockingly, it seems we don't make raw Guzzle requests anywhere else in a way that requires mocking, so this requires adding a new service via DI.

**Alternatives considered**

Keep abusing the Acquia Cloud Client for the subsequent request like before. Doesn't seem smart.

**Testing steps**

1. Follow the [contribution guide](https://github.com/acquia/cli/blob/HEAD/CONTRIBUTING.md#building-and-testing) to set up your development environment.
3. `acli pull:db c[SNIP]b.prod --no-import`
4. Try to `gunzip` the downloaded file. This would be invalid before, but should pass now.

### DIFF
--- a/config/prod/services.yml
+++ b/config/prod/services.yml
@@ -141,3 +141,6 @@ services:
 
   # Amplitude service.
   Zumba\Amplitude\Amplitude: ~
+
+  # Guzzle service.
+  GuzzleHttp\Client: ~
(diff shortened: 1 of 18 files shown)


danepowell:
@fiasco could you please help test since I think you reported this? Here's a dev build: https://transfer.sh/iIHuOx/acli.phar

PR #1191, opened 2022-10-01
BUG-01: Fixes acli remote:aliases:download command.

### LINKED ISSUE #1190: CLI-863: The remote:aliases:download command fails with `unknown drush version` error
When I run `acli remote:aliases:download` command to download latest drush aliases, I am getting an error `Unknown drush version`.

Upon investigating the above issue it was found that  in #1175 [line number 115 ](https://github.com/acquia/cli/pull/1175/files#diff-cd426bdd41a38b7df7a4ca9236f446c31b0f1410c67c6d072c5b12f8caf90cb3L115) the switch statement was updated php8's new `match()` function & as per the  [php documentation for the match](https://www.php.net/manual/en/control-structures.match.php), is an alternative to switch operation but match function `compares values strictly (===) instead of loosely as the switch statement does`. 

Hence due to this strict matching when we pass the drush version argument in the terminal the type of the version number passed is string which is getting compared by the integer which leads to fail the case and hence it default backs to the default condition which raises an error "Unknown Drush Version". 

**Steps to reproduce:**
Step 1: In local env just run `acli remote:aliases:download` in any acquia drupal project.

**Resolution:**

Lets typecast the version parameter that is getting passed to match function.

Tdnshah:
**Motivation**

Fixes #1190

**Proposed changes**

This PR adds an integer type casting the version number passed to match function.

**Alternatives considered**

**Testing steps**

Step 1: In local env just run acli remote:aliases:download in any acquia drupal project.

### DIFF
--- a/src/Command/Remote/AliasesDownloadCommand.php
+++ b/src/Command/Remote/AliasesDownloadCommand.php
@@ -112,9 +112,8 @@ class AliasesDownloadCommand extends SshCommand {
       $this->drushAliasesDir = $this->input->getOption('destination-dir');
     }
     elseif (!isset($this->drushAliasesDir)) {
-      $this->drushAliasesDir = match ($version) {
-        8 => $this->localMachineHelper
-            ->getLocalFilepath('~') . '/.drush',
+      $this->drushAliasesDir = match ((int) $version) {
+        8 => $this->localMachineHelper->getLocalFilepath('~') . '/.drush',
         9 => Path::join($this->getRepoRoot(), 'drush'),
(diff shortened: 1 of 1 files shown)


danepowell:
Fix is good but I'm going to update this PR to fix the test case that should have caught this

danepowell:
Thanks for the contribution, I'm going to clean it up and add tests: https://github.com/acquia/cli/pull/1209

Tdnshah:
No worries. Thank you for the attribution.
This would help me in my upcoming contribution, Like the fixes should be having test cases. :)

PR #1213, opened 2022-10-20
CLI-868: Use file-based refresh tokens

danepowell:


danepowell:
This is the wrong approach. This would require creating a new connector for every request, when connectors already have (or should have) a built-in method to refresh tokens: https://github.com/acquia/cli/pull/1214

PR #1252, opened 2022-12-01
IDE-2744: Add IDE UUID header

danepowell:


anavarre:
@itafroma could you please let us know if those changes would work?

Issue #1297, opened 2023-01-03
CLI-922: [New] Add new ACLI command to change / modify the PHP version for code studio


joshirohit100:
I see people stumbled upon this while PHP version of code studio and project not matching. By default code studio uses PHP 8.0 and if someone creating new project (or updating old one) using php-8.1, face the issue.

Though PHP version can be changed via code studio UI as well as this is JUST a variable but I think would be good if we have something which can be done via CLI (and I don;t need to navigate/click through UI)

PR #1300, opened 2023-01-04
Adding new command to update PHP version for code studio.

joshirohit100:
Issue - https://github.com/acquia/cli/issues/1297
Screenshot 2023-01-04 at 6 31 42 PM

### DIFF
--- a/src/Command/CodeStudio/CodeStudioCommandTrait.php
+++ b/src/Command/CodeStudio/CodeStudioCommandTrait.php
@@ -8,6 +8,7 @@ use Gitlab\Client;
 use Gitlab\Exception\RuntimeException;
 use Gitlab\HttpClient\Builder;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 
 trait CodeStudioCommandTrait {
@@ -271,6 +272,18 @@ trait CodeStudioCommandTrait {
       'topics' => 'Acquia Cloud Application',
       'container_registry_access_level' => 'disabled',
     ];
+  }
+
+  /**
+   * Add gitlab options to the command.
+   *
+   * @return $this
+   */
+  protected function acceptGitlabOptions(): static {
+    $this->addOption('gitlab-token', NULL, InputOption::VALUE_REQUIRED, 'The GitLab personal access token that will be used to communicate with the GitLab instance')
+      ->addOption('gitlab-project-id', NULL, InputOption::VALUE_REQUIRED, 'The project ID (an integer) of the GitLab project to configure.')
+      ->addOption('gitlab-host-name', NULL, InputOption::VALUE_REQUIRED, 'The GitLab hostname.');
+    return $this;
   }
 
 }
--- a/src/Command/CodeStudio/CodeStudioCommandTrait.php
+++ b/src/Command/CodeStudio/CodeStudioCommandTrait.php
@@ -279,7 +279,7 @@ trait CodeStudioCommandTrait {
    *
    * @return $this
    */
-  protected function acceptGitlabOptions(): static {
+  private function acceptGitlabOptions(): static {
     $this->addOption('gitlab-token', NULL, InputOption::VALUE_REQUIRED, 'The GitLab personal access token that will be used to communicate with the GitLab instance')
       ->addOption('gitlab-project-id', NULL, InputOption::VALUE_REQUIRED, 'The project ID (an integer) of the GitLab project to configure.')
       ->addOption('gitlab-host-name', NULL, InputOption::VALUE_REQUIRED, 'The GitLab hostname.');
--- a/src/Command/CodeStudio/CodeStudioPhpVersionCommand.php
+++ b/src/Command/CodeStudio/CodeStudioPhpVersionCommand.php
@@ -50,15 +50,14 @@ class CodeStudioPhpVersionCommand extends CommandBase {
       return self::FAILURE;
     }
 
-    $php_version_already_set = FALSE;
-    // Get all variables of the project.
-    $all_project_variables = $this->gitLabClient->projects()->variables($project['id']);
-    if (!empty($all_project_variables)) {
-      $variables = array_column($all_project_variables, 'value', 'key');
-      $php_version_already_set = $variables['PHP_VERSION'] ?? FALSE;
-    }
-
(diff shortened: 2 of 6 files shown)


joshirohit100:
Added initial stab for review. If look good, will add test case as well

anavarre:
I'm not sure I fully understand the value of this command but what I know is it should be `codestudio:php-version` to have the same command name as `ide:php-version` per https://github.com/acquia/cli/blob/83dfb30525e7f7bea326c3cc53b00566f2474d68/src/Command/Ide/IdePhpVersionCommand.php#L21

Also, we should likely extract some paradigms from this IDE file, such as https://github.com/acquia/cli/blob/83dfb30525e7f7bea326c3cc53b00566f2474d68/src/Command/Ide/IdePhpVersionCommand.php#L87

What are your thoughts @danepowell ?

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+    // If no project found matching with attached Acquia CLoud App id.
+    if (empty($project_id)) {
+      $this->io->error('There are no Acquia Cloud projects available in Code Studio.');
+      return 1;
+    }
+
+    try {
+      // If PHP version is not set in variables.
+      if (!$php_version_already_set) {
+        if ($php_version === '8.0') {
```
danepowell, line 81: Will this _always_ be the default? I doubt it. I'd rather not hardcode this message and have it become stale when the default version changes.

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * @param \Symfony\Component\Console\Input\InputInterface $input
+   * @param \Symfony\Component\Console\Output\OutputInterface $output
+   *
+   * @return int
+   * @throws \Exception
+   */
+  protected function execute(InputInterface $input, OutputInterface $output): int {
+    $php_version = $input->getArgument('php-version');
```
danepowell, line 37: Validate the PHP version. Use https://github.com/acquia/cli/blob/83dfb30525e7f7bea326c3cc53b00566f2474d68/src/Command/Ide/IdePhpVersionCommand.php#L87 . Move that method out of the IdeCommandTrait to somewhere more common like the command base, and pull the last line that's specific to IDEs.

danepowell (changes requested):
This is a great start, but it will also need unit tests.

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+
+  protected static $defaultName = 'codestudio:update-php-version';
+
+  /**
+   * {inheritdoc}.
+   */
+  protected function configure(): void {
+    $this->setDescription('Change the PHP version in Code Studio')
+      ->addArgument('php-version', InputArgument::REQUIRED, 'The PHP version that needs to configured or updated')
+      ->addOption('gitlab-token', NULL, InputOption::VALUE_REQUIRED, 'The GitLab personal access token that will be used to communicate with the GitLab instance')
```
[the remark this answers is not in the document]
joshirohit100: Moved to `CodeStudioCommandTrait` with new method `acceptGitlabOptions()`

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+      ->addOption('gitlab-host-name', NULL, InputOption::VALUE_REQUIRED, 'The GitLab hostname.');
```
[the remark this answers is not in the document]
joshirohit100: Now project id is available as gitlab options moved to common method but not using the projectId in this command. As  fetching the gitlab project based the provided acquia cloud app.

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
```diff
@@ -0,0 +1,102 @@
...
+    $project_id = 0;
+    $php_version_already_set = FALSE;
+    foreach ($ownedProjects as $project) {
+      // Get all the variables set for the project.
+      $all_project_variables = $this->gitLabClient->projects()->variables($project['id']);
+      if (!empty($all_project_variables)) {
+        $variables = array_column($all_project_variables, 'value', 'key');
+        // Assuming this project is attached with Acquia cloud only when
+        // ACQUIA_APPLICATION_UUID variable is available/configured.
+        if (!empty($variables['ACQUIA_APPLICATION_UUID']) && $variables['ACQUIA_APPLICATION_UUID'] === $acquiaCloudAppId) {
```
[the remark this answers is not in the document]
joshirohit100: Removed this and not required now.

src/Command/CodeStudio/CodeStudioChangePHPVersionCommand.php
joshirohit100: removed this.

joshirohit100:
Will work on tests now

joshirohit100:
Added the test cases

joshirohit100 (commented):
All feedbacks addressed

danepowell:
Thanks so much for the hard work, this is 95% good. Can you please just kill a few of the escaped mutants? You don't need to catch them all, although that would be awesome. Usually this just requires adding more assertions to the test cases.

I know not everyone is familiar with mutation testing, if you need more specific guidance let me know.

src/Command/CodeStudio/CodeStudioCommandTrait.php
```diff
@@ -273,4 +274,16 @@ private function getGitLabProjectDefaults(): array {
     ];
   }
 
+  /**
+   * Add gitlab options to the command.
+   *
+   * @return $this
+   */
+  protected function acceptGitlabOptions(): static {
```
danepowell, line 282: This is an easy mutant to kill, just make this method private if it's not being overridden anywhere.

joshirohit100: Fixed!

joshirohit100:
Killed some mutants. There are some still and not sure of how to kill them (Also not sure if need to be killed or not)

danepowell:
Looks good, thanks!

PR #1342, opened 2023-01-27
Adding some test cases for the EnvCronCopy and EnvDelete command.

joshirohit100:
Adding some more tests for the EnvCronCopy and EnvDelete command as when those command were added, I was struggling with writing test cases because of almost no knowledge of testing  :-). Now I have some confidence (though still learning) :)


danepowell:
Awesome, thanks!

PR #1379, opened 2023-02-26
Fix #1374: Adding DB host for code studio

### LINKED ISSUE #1374: CLI-961: acli pull:database command when used in code studio using localhost as DB host
In code studio, the DB host is default to "mysql". refernce -https://code.acquia.com/acquia/standard-template/-/blob/master/gitlab-ci/Auto-DevOps.acquia.gitlab-ci.yml#L20

When this command is used in ACLI (which is being used currently), it suppose to use the mysql as host name or using the correct env variable
 
But https://github.com/acquia/cli/blob/7dacbe94d7007fdf9dab3789119c1e1064d1b7e8/src/Command/CommandBase.php#L244 it seems in order to work this with code studio, this variable has to be override.

So I think there should also be a another check for `DB_HOST` variable

joshirohit100:
Fix https://github.com/acquia/cli/issues/1374

Reference from upstream https://github.com/acquia/drupal-environment-detector/pull/62/files

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -241,6 +241,11 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
     if ($lando_info = self::getLandoInfo()) {
       $this->localDbHost = $lando_info->database->hostnames[0];
     }
+    // If code studio pipeline.
+    else if (!empty(getenv('CI_JOB_ID')) && !empty(getenv('ACQUIA_GLAB_TOKEN_NAME'))) {
+      $this->localDbHost = 'mysql';
+    }
+
     if (getenv('ACLI_DB_HOST')) {
       $this->localDbHost = getenv('ACLI_DB_HOST');
     }
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -242,7 +242,7 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
       $this->localDbHost = $lando_info->database->hostnames[0];
     }
     // If code studio pipeline.
-    else if (!empty(getenv('CI_JOB_ID')) && !empty(getenv('ACQUIA_GLAB_TOKEN_NAME'))) {
+    else if (AcquiaDrupalEnvironmentDetector::isCodeStudioEnv()) {
       $this->localDbHost = 'mysql';
     }
 


src/Command/CommandBase.php
```diff
@@ -241,6 +241,11 @@ private function setLocalDbHost(): void {
     if ($lando_info = self::getLandoInfo()) {
       $this->localDbHost = $lando_info->database->hostnames[0];
     }
+    // If code studio pipeline.
+    else if (!empty(getenv('CI_JOB_ID')) && !empty(getenv('ACQUIA_GLAB_TOKEN_NAME'))) {
+      $this->localDbHost = 'mysql';
+    }
```
joshirohit100, line 247: I tried to add test for this and seems it will be much like `PullDatabaseCommandTest::testPullDatabasesIntoLando()` with only change in mysql host (and env variable set). But then I tried to change some code for testPullDatabasesIntoLando() like changed the db host to `localhost`, also in LandoInfoHelper but looks like tests are always passing. So not sure if relevant to add a test for ths as that will be just a copy paste

joshirohit100:
Given https://github.com/acquia/drupal-environment-detector/pull/62 is merged, we have to wait once that is available here and then PR can be updated using the CS change

danepowell:
I published a new release of the environment detector: https://github.com/acquia/drupal-environment-detector/releases/tag/1.6.0

src/Command/CommandBase.php
```diff
@@ -241,6 +241,11 @@ private function setLocalDbHost(): void {
     if ($lando_info = self::getLandoInfo()) {
       $this->localDbHost = $lando_info->database->hostnames[0];
     }
+    // If code studio pipeline.
+    else if (AcquiaDrupalEnvironmentDetector::isCodeStudioEnv()) {
```
joshirohit100, line 245: Updated the acquia/drupal-environment-detector to 1.6.0

joshirohit100:
Updated AcquiaDrupalEnvironmentDetector library as well in this pr to 1.6.0

One thing I noticed, may be already known and expected - While updating the package via composer, I had to create and use the GH token because of `Your GitHub credentials are required to fetch private repository metadata (https://github.com/grasmash/dns-zone-configurator)` as seems considering this as private or something

danepowell:
We had to fork dns-zone-configurator as it appears abandoned. I think Composer treats any custom repo as private even if it's public, which is a little confusing. I created an internal issue to unfork that if possible.

danepowell:
I see an internal MR has been raised to set ACLI_DB_HOST, that seems like the better solution.

Issue #1362, opened 2023-02-15
CLI-958: New command to delete the users from organnization based on email address


joshirohit100:
We have existing ACLI command to delete the user from organization -
```
acli api:organizations:member-delete --help
Description:
  Removes the member from the organization.

Usage:
  api:organizations:member-delete <organizationUuid> <userUuid>
  api:organizations:member-delete 1e7efab9-0fac-4a2c-ad94-61efc78623ba f2daa9cc-e5a0-4036-a5c8-f96e336c62b5
```

But this command takes uses the userID.

So what If I have list of email addresses (multiple) and want to delete from the given organization?

cc @milind-kagdelwar

PR #1389, opened 2023-03-06
Adding new command for removing members from organization using email…

joshirohit100:
For the issue https://github.com/acquia/cli/issues/1362

```
➜ ./bin/acli org:member-remove --help
Description:
  Remove the members from an organiztaion on Acquia Cloud Platform

Usage:
  organization:member-remove <organization_id> <user_emails>
  organization:member-remove 1e7efab9-0fac-4a2c-ad94-61efc78623ba mytest@email.com
  organization:member-remove 1e7efab9-0fac-4a2c-ad94-61efc78623ba mytest1@email.com,mytest2@email.com

Arguments:
  organization_id       ID of the organization
  user_emails           Comma separated emails of the users to be removed from organization

Options:
  -h, --help            Display help for the given command. When no command is given display help for the list command
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi|--no-ansi  Force (or disable --no-ansi) ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug
```

cc @milind-kagdelwar


joshirohit100:
Once reviewed and looks fine this command, I will add tests

tests/phpunit/src/Application/KernelTest.php
```diff
@@ -77,6 +77,8 @@ private function getEnd(): string {
   ide:list:app             [ide:list] List available Cloud IDEs belonging to a given application
   ide:list:mine            List Cloud IDEs belonging to you
   ide:open                 Open a Cloud IDE in your browser
+ org
+  org:member-remove        Remove the members from an organiztaion on Acquia Cloud Platform
```
joshirohit100, line 81: Because of this test failure, I changed `organization:member-remove` => `org:member-remove` or update this complete test file (spaces :) )

danepowell:
It seems like this is primarily a wrapper for the postOrganizationMemberDelete endpoint that translates email addresses to user UUIDs. Do other endpoints accept a user UUID?

If so, I think a better solution would be to automatically translate emails to UUIDs for endpoints that accept UUIDs.

This is a pattern we already follow for application and environment aliases.

joshirohit100:
Yes! its kind of wrapper.
 
So for deleting multiple accounts at once, the customer/user has to create some wrapper script to loop through email addresses and execute this command?

danepowell:
This would be the first ACLI command to accept a list as an argument. I'm not saying that's necessarily bad, but it makes me suspicious. Is removing multiple members at once from an organization really so common?

I'd really prefer to see a more general solution to translate email addresses to user ids. Basically, add a new method `convertUserAliasToUuid()` here: https://github.com/acquia/cli/blob/7d23270e75bf5568b3ecfb418e076b2d6b616a6f/src/Command/CommandBase.php#L225

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
...
+   *   User alias like uuid or email.
+   * @param String $orgUuidArgument
+   *   Organization uuid.
+   * @return string
+   *   User uuid from alias
+   */
+  protected function getUserUuidFromUserAlias(String $userAlias, String $orgUuidArgument): string {
+    $acquiaCloudClient = $this->cloudApiClientService->getClient();
+    $organization_resource = new Organizations($acquiaCloudClient);
+    $org_members = $organization_resource->getMembers($orgUuidArgument);
```
joshirohit100, line 1076: Organization api also separate method to get the admins. For now, added only members.
If required, we can also get admins and merge the list to check from both

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -400,4 +400,60 @@ public function testGetApplicationUuidFromBltYml(): void {
...
+   * Test of deletion of the user from organization by user email.
+   */
+  public function testOrganizationMemberDeleteByUserEmail(): void {
+    ClearCacheCommand::clearCaches();
+    $org_id = 'bfafd31a-83a6-4257-b0ec-afdeff83117a';
+    $member_uuid = 'james.kirk@example.com';
+
+    $members_response = $this->getMockResponseFromSpec('/organizations/{organizationUuid}/members', 'get', 200);
+    $this->clientProphecy->request('get', '/organizations/' . $org_id . '/members')
+      ->willReturn($members_response->_embedded->items)->shouldBeCalled();
```
joshirohit100, line 439: couldn;t user the data provider because of this extra api call and then use of `$members_response->_embedded->items[0]->uuid` below

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
     }
   }
 
+  /**
+   * @param $userUuidArgument
+   * @param $orgUuidArgument
+   */
+  protected function convertUserAliasToUuid(InputInterface $input, $userUuidArgument, $orgUuidArgument): void {
```
danepowell, line 1032:
```suggestion
  private function convertUserAliasToUuid(InputInterface $input, string $userUuidArgument, string $orgUuidArgument): void {
```

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
     }
   }
 
+  /**
+   * @param $userUuidArgument
+   * @param $orgUuidArgument
+   */
```
danepowell, line 1031:
```suggestion
```

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
...
+
+  /**
+   * @param String $userAlias
+   *   User alias like uuid or email.
+   * @param String $orgUuidArgument
+   *   Organization uuid.
+   * @return string
+   *   User uuid from alias
+   */
+  protected function validateUserUuid($userUuidArgument, $orgUuidArgument): mixed {
```
danepowell, line 1053:
```suggestion
  private function validateUserUuid(string $userUuidArgument, string $orgUuidArgument): string {
```

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
...
+
+  /**
+   * @param String $userAlias
+   *   User alias like uuid or email.
+   * @param String $orgUuidArgument
+   *   Organization uuid.
+   * @return string
+   *   User uuid from alias
+   */
+  protected function getUserUuidFromUserAlias(String $userAlias, String $orgUuidArgument): string {
```
danepowell, line 1073:
```suggestion
  private function getUserUuidFromUserAlias(string $userAlias, string $orgUuidArgument): string {
```

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -400,4 +400,60 @@ public function testGetApplicationUuidFromBltYml(): void {
     $this->fs->remove($bltConfigFilePath);
   }
 
+  /**
+   * Test of deletion of the user from organization by user uuid.
+   */
+  public function testOrganizationMemberDeleteByUserUuid(): void {
+    ClearCacheCommand::clearCaches();
```
danepowell, line 407: Is clearing caches absolutely necessary? Clearing caches is not thread safe, so if you really need this you need to add this test to the serial group. I think the annotation is `@group serial`. But in most cases, you don't need to clear caches as part of a test.

src/Command/CommandBase.php
```diff
@@ -1023,6 +1025,71 @@ protected function fillMissingRequiredApplicationUuid(InputInterface $input, Out
...
+    $acquiaCloudClient = $this->cloudApiClientService->getClient();
+    $organization_resource = new Organizations($acquiaCloudClient);
```
danepowell, line 1075: Please convert all variables to camelCase

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -400,4 +400,60 @@ public function testGetApplicationUuidFromBltYml(): void {
...
+    $org_id = 'bfafd31a-83a6-4257-b0ec-afdeff83117a';
+    $member_uuid = '26c4af83-545b-45cb-b165-d537adc9e0b4';
+
+    $response = $this->getMockResponseFromSpec('/organizations/{organizationUuid}/members/{userUuid}', 'delete', 200);
```
danepowell, line 411: Note to self, I'll try to replace this with the new mockResponse method (may require some tweaks to that method)

danepowell (changes requested):
This is looking great, mainly needs code style cleanup

src/Command/CommandBase.php
joshirohit100: Fixed!

src/Command/CommandBase.php
joshirohit100: fixed!

src/Command/CommandBase.php
joshirohit100: fixed!

src/Command/CommandBase.php
joshirohit100: fixed!

tests/phpunit/src/Commands/Api/ApiCommandTest.php
joshirohit100: removed!

danepowell (approved):
Thanks for doing the legwork on this!

PR #1396, opened 2023-03-08
CLI-973: Track SSH key upload time

danepowell:


src/Command/Ssh/SshKeyCommandBase.php
```diff
@@ -183,8 +189,10 @@ static function () use ($spinner) {
     // Poll Cloud every 5 seconds.
     $timers[] = $loop->addPeriodicTimer(5, $callback);
     $timers[] = $loop->addTimer(0.1, $callback);
-    $timers[] = $loop->addTimer(10 * 60, function () use ($output, $loop, &$timers) {
+    $timers[] = $loop->addTimer(60 * 60, function () use ($output, $loop, &$timers) {
```
danepowell, line 192: @anavarre I'm adding the SSH uploading timing analytics and cutting a new release with just regression testing, even though I can't see the results in Amplitude, because I _really_ want to get this info as soon as possible.

I also extended the timeout from 10 to 60 minutes so that we get more granular data than just "more than 10 minutes", which I think is often the case. If you disagree let me know tomorrow and I'll fix it in a new release.

anavarre: I think a longer timeout is good.

PR #1424, opened 2023-03-14
CLI-948: Consistent option filling

danepowell:
**Motivation**

When a command requires certain options but they are not provided as arguments, we must prompt for them. We do this a thousand different ways across all commands, leading to an inconsistent user experience.

Screenshot 2023-03-14 at 2 23 06 PM
Screenshot 2023-03-14 at 2 23 18 PM

**Proposed changes**

Use a single consistent method of asking for required options.

Screenshot 2023-03-14 at 2 24 25 PM
Screenshot 2023-03-14 at 2 24 46 PM

@anavarre do you like this UX? The goal is to keep it consistent across commands, and to promote the CLI options which the user may not notice otherwise. Ignore the errors, they aren't related.

**Testing steps**


anavarre:
I think it makes sense to make the UX consistent. There's nothing more confusing than having different patterns. I certainly am not a fan of having a note with exclamation marks to somewhat use different words to say the same thing than the user prompt.

With that being said, I think we need to be careful to not remove some meaning from the consolidation. As an example, in the initial screen where you enter your factory URL, we're showing you the example of `https://www.acquia.com` which is good to know about. In the new screen I know I have to enter the URL but have no idea whether I'm going to fail validation if I enter the protocol or omit it.

danepowell:
> initial screen where you enter your factory URL, we're showing you the example of https://www.acquia.com

Incidentally, that's a really poor example because that would never be a factory URL, but I take your point :) If the option provides example usage, we should provide it in the question.

anavarre:
Yes, I might have poorly indicated that we want to provide users with a way to know what is expected for them to fill in and have validation to have their back with more info.

danepowell:
Lol, you didn't do anything poorly. I was already working on fixing that example and found it amusing that you picked it out, is all.

danepowell:
It's cumbersome to provide examples (like your API key) in the help text and input prompt, since the API key alone takes up half the terminal. And it's not appropriate to provide these as defaults, since IMHO defaults should only contain valid data (not something like a dummy API key).

I like the current pattern, which only provides example data as part of the validator if the input fails.

danepowell:
I've implemented this for `auth:login`, `auth:acsf-login`, and `ide:create`. Let's see if we get any feedback on those before rolling it out to other commands.

PR #1444, opened 2023-03-22
CLI-996: [env:certificate-create] new command

danepowell:
**Motivation**

The existing `api:environments:certificate-create` command is difficult to use because you have to pass certificates as arguments, which can be thousands of characters, and the help text is incorrect and certificate `---` lines are interpreted as Bash argument separators.

**Proposed changes**

Add a new command `env:certificate-create` that accepts filenames of certicates.

**Alternatives considered**

Change help text on `api:environments:certificate-create` to show people how to simply read a file as an argument. But the existing help text would still be broken.

**Testing steps**

3. `wacli env:certificate-create ~/Downloads/cert.pem ~/Downloads/key.pem foo.dev --ca_certificates=~/Downloads/ca.pem`

### DIFF
--- a/src/Command/Env/EnvCertCreateCommand.php
+++ b/src/Command/Env/EnvCertCreateCommand.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Acquia\Cli\Command\Env;
+
+use Acquia\Cli\Command\CommandBase;
+use AcquiaCloudApi\Endpoints\SslCertificates;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Class EnvCertCreateCommand.
+ */
+class EnvCertCreateCommand extends CommandBase {
+
+  protected static $defaultName = 'env:certificate-create';
+
+  /**
+   * {inheritdoc}.
+   */
+  protected function configure(): void {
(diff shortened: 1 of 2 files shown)


src/Command/Env/EnvCertCreateCommand.php
```diff
@@ -0,0 +1,54 @@
...
+    $sslCertificates->create(
+      $environment_uuid,
+      $input->getOption('label'),
+      $this->localMachineHelper->readFile($input->getArgument('certificate')),
+      $this->localMachineHelper->readFile($input->getArgument('private_key')),
+      $this->localMachineHelper->readFile($input->getOption('ca_certificates')),
+      $input->getOption('csr_id'),
+      $input->getOption('legacy')
+    );
+    $this->io->success('Certificate was installed');
```
eporama, line 50: Should probably changed this to "Certificate being installed." as this creates the task, but can we get the notification link and then run through `application:task-wait` machinations to make sure it is completely finished?

danepowell:
@eporama or @anavarre are you able to help test this? I'll provide example certs in the Jira ticket it you need.

The certificate and private key are required parameters, all of the other params are optional. We prompt for those optional params if they are missing. I'm not sure if that's a good or bad pattern to establish 😕 

Also the API command would require the certificate and private key as unnamed arguments rather than options, but I kind of hate this... I feel like everything should be an option (not an argument) so it can be explicitly named. Hence why they are options here.

danepowell:
Changed my mind and made the required options argument instead.

PR #1446, opened 2023-03-23
CLI-1004: BugSnag reports from downstream projects

### LINKED ISSUE #1445: CLI-1004: BugSnag reports from downstream projects
Projects that consume Acquia CLI as a library (which is _super_ unsupported, by the way) are throwing errors that get caught by our BugSnag instance.

I think we should set the BugSnag token in `bin/acli` so that it only gets picked up during CLI execution. Or better still, inject it at build time so it's only present in the Phar.

danepowell:
Fix #1445 

Wish we had https://github.com/box-project/box/issues/890


.github/workflows/ci.yml
```diff
@@ -82,7 +82,6 @@ jobs:
           composer box-install
           # Warm the symfony cache so it gets bundled with phar.
           ./bin/acli
-          composer dump-env prod
```
danepowell, line 85: Turns out this does nothing because we don't use the framework bundle. Can't load .env files, etc.

PR #1473, opened 2023-04-18
CLI-1036: [codestudio:php-version] null applicationUuid

danepowell:
`CommandBase::getCloudApplication(): Argument #1 ($application_uuid) must be of type string, null given`

### DIFF
--- a/src/Command/CodeStudio/CodeStudioPhpVersionCommand.php
+++ b/src/Command/CodeStudio/CodeStudioPhpVersionCommand.php
@@ -27,14 +27,13 @@ class CodeStudioPhpVersionCommand extends CommandBase {
     $php_version = $input->getArgument('php-version');
     $this->validatePhpVersion($php_version);
     $this->authenticateWithGitLab();
-    $acquiaCloudAppId = $input->getArgument('applicationUuid');
+    $acquiaCloudAppId = $this->determineCloudApplication();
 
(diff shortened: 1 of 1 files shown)


src/Command/CodeStudio/CodeStudioPhpVersionCommand.php
```diff
@@ -27,14 +27,13 @@ protected function execute(InputInterface $input, OutputInterface $output): int
     $php_version = $input->getArgument('php-version');
     $this->validatePhpVersion($php_version);
     $this->authenticateWithGitLab();
-    $acquiaCloudAppId = $input->getArgument('applicationUuid');
+    $acquiaCloudAppId = $this->determineCloudApplication();
```
danepowell, line 30: @joshirohit100 `determineCloudApplication()` is the canonical way of getting an application ID. It's better than just getting the `applicationUuid` argument because it allows getting the application in other ways (datastore, Git config), and guarantees that the id is not null.

Let know if you think this change will cause any issues.

joshirohit100: I think tits fine (better :) ) as same is being done at https://github.com/acquia/cli/blob/1d629e775d9ff9ceb9565910ad1f4598380c14cf/src/Command/CommandBase.php#L800

PR #1498, opened 2023-05-15
CLI-1053: Convert snake case to camel case

danepowell:
I'd like to be able to add this to our phpcs standard:

```
  <rule ref="Squiz.NamingConventions.ValidVariableName"/>
  <rule ref="Squiz.NamingConventions.ValidVariableName.PrivateNoUnderscore">
    <severity>0</severity>
  </rule>
```

Unfortunately because upstream libraries use snake case, any references to those objects and methods get flagged as false positives, and I don't know how to exclude them en masse.


danepowell:
In hindsight, I should have done the src and tests directories as separate commits to prevent common mode failures like #1504

PR #1510, opened 2023-05-22
Refactor mock methods

danepowell:
The mock methods are basically all boilerplate, and they are a bit magical and inconsistent in how they handle requests, parameters, and responses.

It might be a little more code, but I'd prefer to reduce the amount of abstraction and have just a single mockRequest method.


danepowell:
I don't have the time or patience to refactor the whole codebase right now but I like this new pattern.

PR #1515, opened 2023-05-31
CLI-1054: Support pull commands outside of IDEs

### LINKED ISSUE #1386: CLI-971: Pull command is not documented
`acli pull` is a very useful command, however it is not documented [here](https://docs.acquia.com/acquia-cli/commands/). Also, when I run `acli` in my terminal I do not see the pull command listed. This issue occurs on both the latest 1.30.1 branch and 2.8.1

I would expect to see this command listed in both places

danepowell:
Fix #1386 

@mikemadison13 @anavarre any opinions about this as a solution to avoiding confusion about documented commands?

Most of the hidden commands (i.e., `pull:files`) were only hidden because they required a db connection that was only set in an IDE. But we now have environment variables like ACLI_DB_HOST that anyone could set locally if they really wanted.

This unhides those commands and instead advises via help text that these variables should be set:
Screenshot 2023-05-31 at 10 50 51 AM

We still hide truly IDE-specific commands (i.e., `ide:service-restart`). Hopefully the reason is obvious enough that people aren't still confused.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -10,7 +10,6 @@
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
         convertDeprecationsToExceptions="true"
-        bootstrap="tests/phpunit/bootstrap.php"
 >
   <php>
     <env name="AMPLITUDE_KEY" value=""/>
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -85,7 +85,6 @@ abstract class CommandBase extends Command implements LoggerAwareInterface {
   protected string $localDbPassword = 'drupal';
   protected string $localDbName = 'drupal';
   protected string $localDbHost = 'localhost';
-
   protected bool $drushHasActiveDatabaseConnection;
   protected \GuzzleHttp\Client $updateClient;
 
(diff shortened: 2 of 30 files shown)


src/Command/CommandBase.php
```diff
@@ -109,6 +108,19 @@ public function __construct(
     $this->setLocalDbName();
     $this->setLocalDbHost();
     parent::__construct();
+    if ($this->commandRequiresAuthentication()) {
+      $this->appendHelp('This command requires authentication via the Cloud Platform API.');
+    }
+    if ($this->commandRequiresDatabase()) {
+      $this->appendHelp('This command requires an active database connection. Set the following environment variables prior to running this command: '
+        . 'ACLI_DB_HOST, ACLI_DB_NAME, ACLI_DB_USER, ACLI_DB_PASSWORD');
```
anavarre, line 116: This could be the compromise we want. But aren't we going to get support tickets on how to set `ACLI_DB_HOST, ACLI_DB_NAME, ACLI_DB_USER, ACLI_DB_PASSWORD`? Should we further clarify that the command shouldn't typically be used outside of an IDE?

phpunit.xml.dist
```diff
@@ -10,7 +10,6 @@
         convertNoticesToExceptions="true"
         convertWarningsToExceptions="true"
         convertDeprecationsToExceptions="true"
-        bootstrap="tests/phpunit/bootstrap.php"
```
danepowell, line 13: I wasn't able to recreate an instance where clearing cache was actually necessary to pick up changes in command files.

PR #1527, opened 2023-06-07
Throw exception if userAlias not found

danepowell:
Right now if you pass an email address as user alias and that email doesn't match someone in the org you get a confusing error message from the API:

```
{
    "error": "invalid_id",
    "message": "\"somebody@danepowell.com\" is not a valid v4 UUID."
}
```

ACLI should not pass an argument to the API that it knows to be invalid (i.e., not a UUID).

@joshirohit100 would you mind reviewing since you wrote this?


joshirohit100:
Looks good to me!

PR #1536, opened 2023-06-09
Slow windows tests

danepowell:
Windows tests runs are brutally slow compared to linux. Trying to figure out why.

- https://github.com/actions/runner-images/issues/7320
- https://github.com/actions/runner-images/issues/5166

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,10 @@ jobs:
             php: "8.2"
             coverage: "none"
     steps:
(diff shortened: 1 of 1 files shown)


danepowell:
I think https://github.com/actions/runner-images/issues/7320 is the culprit. Any CPU-intensive task (composer install, tests) just takes forever. It might be a noisy-neighbor problem, as our midnight cron jobs never take long at all, but PR tests during the day can take 2-5x longer than Ubuntu.

PR #1567, opened 2023-07-10
CLI-1103: Fix #1558: Notification arg formats

### LINKED ISSUE #1558: CLI-1103: Support notification URLs/JSON strings as input for commands accepting a notification UUID
**Is your feature request related to a problem? Please describe.**

I have found that some acli commands return a JSON object with a `notification` property that just has the notification UUID (e.g., `api:environments:code-switch`) while others only return the notification as a URL in the `_links->notification->href` property (e.g., `api:environments:database-copy`).

The `app:task-wait` command supports both the JSON object and plain UUID, which is very helpful. The `api:notifications:find` command only supports a plain UUID.

What I am looking to do is:

1. Execute a task.
2. Use the JSON object response to wait for the task to complete.
3. Check for a `completed` task status.

*Note: this is somewhat related to #1556 which would probably make this use case moot however I do think this feature request still has merit on it's own*

Steps 1 and 2 above are simple enough with `app:task-wait`'s support for the JSON object response. But Step 3 is a little tougher because in some cases I'll need to do some extra work to parse the URL from the `_links->notification->href` property (not super fun in bash, at least).

**Describe the solution you'd like**

I would like for the `app:task-wait` and `api:notifications:find` commands to support any of the following as input:

1. Notification UUID
2. Notification URL
3. JSON object with a `_links->notification->href` property

There may also be other commands that accept a notification UUID where this could be useful but these two are the primary use case I have come up against.

If supporting the notification URL is seen as more out-of-scope, at least bringing the supported input for the commands to parity (so either a plain UUID or a JSON object for `api:notifications:find`) would be ideal.

**Describe alternatives you've considered**

The commands are functional as-is and there are alternatives (like parsing the notification URL) that work fine. This would be a nice-to-have.

danepowell:
**Motivation**

Fixes #1558

**Proposed changes**

Anywhere that a notification UUID is required as argument, support passing a notification object or URL as well. This basically generalizes the `getNotificationUuid` logic in app:task-wait to all commands.

**Testing steps**

3. Try `acli app:task-wait` with various forms of arguments to ensure they still work:

```
# Notifications object response
wacli app:task-wait "$(acli api:environments:domain-clear-caches pipelinesvalidation2.dev pipelinesvalidation2dev.prod.acquia-sites.com)"
# Notification UUID
wacli app:task-wait 227eccb1-3345-4a2d-b042-eddd56a78ed7
# Notification URL
wacli app:task-wait https:\/\/cloud.acquia.com\/api\/notifications\/64df4412-20fa-4489-8cbf-1e73095d03eb
```

@cdubz could you please help test and let me know if it satisfies your request? If you follow the steps above, you can find a pre-built version of CLI for this PR to test with, or you can build it yourself.

### DIFF
--- a/src/Command/App/TaskWaitCommand.php
+++ b/src/Command/App/TaskWaitCommand.php
@@ -5,7 +5,6 @@ declare(strict_types = 1);
 namespace Acquia\Cli\Command\App;
 
 use Acquia\Cli\Command\CommandBase;
-use Acquia\Cli\Exception\AcquiaCliException;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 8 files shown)


cdubz:
@danepowell thanks for picking this up! This does exactly what I'm looking for and will eliminate a lot of complexity in my CI processes (:

I think there is just one small issue to address -- if I give `task-wait` an input that is not a valid UUID or notification URL there will be a PHP fatal error in most cases because the code will get all the way to the URL check where it checks for `$urlParts[5]` without ensuring that array index exists. So we get:

```
PHP Warning:  Undefined array key 5 in phar:///src/Command/CommandBase.php on line 1068
PHP Fatal error:  Uncaught TypeError: Acquia\Cli\Command\CommandBase::validateUuid(): Argument #1 ($uuid) must be of type string, null given, called in phar:///src/Command/CommandBase.php on line 1069 and defined in phar:///src/Command/CommandBase.php:726
```

If that can be avoided it will just end up at the generic "Notification format is not one of UUID, JSON response, or URL" which seems fine.

danepowell:
Thanks for the feedback! I fixed that and a few other bugs. Feel free to hammer on the new version, otherwise I'll merge this after getting coverage up.

PR #1581, opened 2023-08-22
CLI-1121: Fix sql imports

danepowell:
**Motivation**

Drush SQL imports can fail if the SQL file is larger than roughly 64k. For example: `acli drush site.env -- sqlc < big-sql-dump.sql`

This is due to an upstream Symfony bug: https://github.com/symfony/symfony/issues/21580

**Proposed changes**

Implement the same workaround as https://github.com/consolidation/site-process/issues/74 and https://github.com/acquia/mc-cs-kafka-consumer/commit/acca146d7736f80add46c9ba5864a0b53758dc71

I'm not confident this is a good workaround. While it fixes this problem, it may introduce others (as evidenced by the failing tests). For instance, `acli drush site.env -- sql-dump` produces a zero-byte export. Which is weird, because without the redirect it does print to stdout as expected.


danepowell:
The latest commit helps but breaks other things, including checklists (in push:artifact, for instance).

src/Helpers/LocalMachineHelper.php
```diff
@@ -124,14 +122,18 @@ private function configureProcess(Process $process, string $cwd = NULL, ?bool $p
...
         $this->output->write($buffer);
       };
     }
-    $process->start();
-    $process->wait($callback);
+    if ($process->getInput()) {
+      $this->runAsync($process);
+    }
+    else {
+      $process->run($callback);
```
danepowell, line 135: I need to examine the code path for push:artifact and similar commands. If it's runAsync, I'm not surprised it's broken since we're missing the callback. But even adding the callback didn't fix it.

It looks like starting with the callback should be the same as waiting with it. I'm not sure why we did it the way we did initially.

danepowell: Ah, it actually appears that the Symfony 6 upgrade caused the side effects. This code might work fine, though we probably need to pass the callback to run(), and generally make this more consistent.

danepowell:
I just realized, it may not be this simple. I think this actually needs to be fixed in Drush since ACLI is calling Drush. We'd only need to "fix" this in ACLI if some other process was calling ACLI and experiencing the bug.

May need to get traction on https://github.com/consolidation/site-process/issues/74

danepowell:
Now I understand why this _appeared_ to work initially. The replacement error handler was swallowing the pipe error, and the lack of a callback was suppressing the SQL import error. But in fact, nothing was ever fix here. I think this needs to be fix upstream in Symfony and/or site-process and/or Drush first.

PR #1583, opened 2023-08-23
CLI-1133: FedAuth API calls require org scope

danepowell:
Requires https://github.com/typhonius/acquia-php-sdk-v2/pull/418

Obviously this needs work to add documentation and fix the scenario where org id is not set.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,12 +10,6 @@
             "role": "Maintainer"
         }
     ],
-    "repositories": {
-      "sdk": {
-        "type": "path",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -16,6 +16,7 @@
         "ext-json": "*",
         "acquia/drupal-environment-detector": "^1.2.0",
         "bugsnag/bugsnag": "^3.0",
+        "composer/composer": "2.5.8",
```
danepowell, line 19: Necessary to work around an issue where Composer is choosing a dev version of Composer in order to bump the major version of some other dependency

Issue #754, opened 2021-12-14
when run via acli, drush with @self alias does not work


toddzebert:
`acli drush config:pull @my-site.dev @self`

returns:

> Environment alias must match the pattern [app-name].[env]

grasmash:
Run the command using this format:
```
acli drush @my-site.dev -- config:pull @my-site.dev @self
```

grasmash:
Admittedly I haven't used `config:pull` specifically. I wonder how that works.

grasmash:
Actually, that won't work. When you use `acli drush` it runs the command on the remote machine. You want to run the command on the local machine. You really just want to use drush.

grasmash:
Tested. Worked well for me.

PR #1595, opened 2023-09-21
Explicitly set php version to be used during CI build

balsama:
**Motivation**
There were a couple of changes in shivammathur/setup-php (specifically [#754](https://github.com/shivammathur/setup-php/commit/a5fb328c6a38ac8086676fc64e6e78a39d6ef69e)) which resulted in new builds using [PHP 8.0 during the build](https://github.com/acquia/cli/actions/runs/6263194877/job/17007030360#step:6:379) step.

**Proposed changes**
This change explicitly sets the php version to be used.

**Alternatives considered**
None

**Testing steps**
The build step will fail (example) without this with:
```
"Error: ] Your system is not ready to run the application.                       
                                                                                

Fix the following mandatory requirements:
=========================================

 * The application requires a version matching "^8.1"."
 ``` 
 
And it will pass with this PR merged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,7 @@ jobs:
       - uses: shivammathur/setup-php@v2
         with:
           coverage: none
+          php-version: '8.1'
       - name: 'Get ACLI version'
         id: acli-version
         run: |


danepowell:
Thanks. I was initially confused by this report because Acquia CLI itself only requires PHP 8.0. But the Box compiler requires PHP 8.1; that's what was throwing the error when the setup-php action changed and started running PHP 8.0.

This seems like the best solution, short of a separate PHP version for require and require-dev, which is a whole other kettle of fish: https://github.com/composer/composer/issues/5136

PR #1606, opened 2023-10-18
CLI-1136: [push:artifact] include empty directories and symlinks

danepowell:
**Motivation**

push:artifact doesn't mirror symlinks that point to directories. It probably also doesn't mirror empty directories, though that's less likely to cause problems since empty directories are VCS ignored and not generally useful anyway.


mglaman:
@danepowell this is causing issues if a directory existed in the original artifact but was removed in the new artifact.

```
[2023-10-25T16:22:55.449Z]   [Symfony\Component\Finder\Exception\AccessDeniedException]                   
[2023-10-25T16:22:55.450Z]   RecursiveDirectoryIterator::__construct(/tmp/acli-push-artifact/acquia-util  
[2023-10-25T16:22:55.450Z]   s): Failed to open directory: No such file or directory
```

danepowell:
Can you please open a new ticket (GitHub or Jira) with steps to reproduce? I don't understand why removing a directory would be any different than removing a file as far as the Symfony Finder is concerned.

mglaman:
Yeah, I'll open a Jira ticket so I can explain in detail more

PR #1614, opened 2023-10-25
CLI-1181: [pull:code] disable scripts for build branches

danepowell:
**Motivation**

For certain applications and environments, `composer install` can fail when composer dependencies are already installed

**Proposed changes**

Don't run composer install if dependencies are already installed; i.e., for build branches.

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -508,15 +508,15 @@ abstract class PullCommandBase extends CommandBase {
 
   protected function runComposerScripts(callable $outputCallback = NULL): void {
     if (!file_exists(Path::join($this->dir, 'composer.json'))) {
-      $this->logger->notice('composer.json file not found. Skipping composer install.');
+      $this->io->note('composer.json file not found. Skipping composer install.');
       return;
     }
     if (!$this->localMachineHelper->commandExists('composer')) {
-      $this->logger->notice('Composer not found. Skipping composer install.');
+      $this->io->note('Composer not found. Skipping composer install.');
       return;
     }
     if (file_exists(Path::join($this->dir, 'vendor'))) {
-      $this->logger->notice('Composer dependencies already installed. Skipping composer install.');
+      $this->io->note('Composer dependencies already installed. Skipping composer install.');
       return;
     }
     $this->checklist->addItem("Installing Composer dependencies");
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -507,14 +507,21 @@ abstract class PullCommandBase extends CommandBase {
   }
 
   protected function runComposerScripts(callable $outputCallback = NULL): void {
-    if (file_exists($this->dir . '/composer.json') && $this->localMachineHelper->commandExists('composer')) {
-      $this->checklist->addItem("Installing Composer dependencies");
-      $this->composerInstall($outputCallback);
-      $this->checklist->completePreviousItem();
+    if (!file_exists(Path::join($this->dir, 'composer.json'))) {
+      $this->logger->notice('composer.json file not found. Skipping composer install.');
+      return;
     }
-    else {
-      $this->logger->notice('composer or composer.json file not found. Skipping composer install.');
+    if (!$this->localMachineHelper->commandExists('composer')) {
+      $this->logger->notice('Composer not found. Skipping composer install.');
+      return;
     }
+    if (file_exists(Path::join($this->dir, 'vendor'))) {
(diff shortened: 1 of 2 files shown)


src/Command/Pull/PullCommandBase.php
```diff
@@ -507,14 +507,21 @@ private function promptChooseDatabases(
   }
 
   protected function runComposerScripts(callable $outputCallback = NULL): void {
-    if (file_exists($this->dir . '/composer.json') && $this->localMachineHelper->commandExists('composer')) {
-      $this->checklist->addItem("Installing Composer dependencies");
-      $this->composerInstall($outputCallback);
-      $this->checklist->completePreviousItem();
+    if (!file_exists(Path::join($this->dir, 'composer.json'))) {
+      $this->logger->notice('composer.json file not found. Skipping composer install.');
+      return;
```
kaynenh, line 512: Negative exists logic here seems to make sense, I think.

src/Command/Pull/PullCommandBase.php
```diff
@@ -507,14 +507,21 @@ private function promptChooseDatabases(
   }
 
   protected function runComposerScripts(callable $outputCallback = NULL): void {
-    if (file_exists($this->dir . '/composer.json') && $this->localMachineHelper->commandExists('composer')) {
-      $this->checklist->addItem("Installing Composer dependencies");
-      $this->composerInstall($outputCallback);
-      $this->checklist->completePreviousItem();
+    if (!file_exists(Path::join($this->dir, 'composer.json'))) {
+      $this->logger->notice('composer.json file not found. Skipping composer install.');
```
anavarre, line 511: I'm not sure that's the behavior we want. The codebase will have `composer.json` and `composer.lock` + the `vendor` dir. That's the issue.

src/Command/Pull/PullCommandBase.php
```diff
@@ -507,14 +507,21 @@ private function promptChooseDatabases(
...
+      $this->logger->notice('Composer not found. Skipping composer install.');
+      return;
     }
+    if (file_exists(Path::join($this->dir, 'vendor'))) {
+      $this->logger->notice('Composer dependencies already installed. Skipping composer install.');
+      return;
+    }
+    $this->checklist->addItem("Installing Composer dependencies");
+    $this->composerInstall($outputCallback);
+    $this->checklist->completePreviousItem();
```
kaynenh, line 524: Not sure if the mutation tests are catching something here or if they just need to be updated. It seems it's the latter.

kaynenh (approved):
I think I get the logic here but best if you get another approval.

src/Command/Pull/PullCommandBase.php
anavarre: See details at https://acquia.atlassian.net/browse/CLI-1181?focusedCommentId=7939845

danepowell: Right, that check for `vendor` comes later. The check for `composer.json` was from before all this and it's something we want to preserve.

anavarre: Removed my comment. That's a public repo 🤦

src/Command/Pull/PullCommandBase.php
danepowell: I like to do any validation at the start of the method, hence the negative logic. I'm open to alternatives.

src/Command/Pull/PullCommandBase.php
danepowell: Due to idiosyncrasies with how the Symfony Command Tester class handles output, our test cases can't verify that these checklist items appear in the command output. Hence the escaped mutant. But it's not caused by this PR, so we can ignore it.

src/Command/Pull/PullCommandBase.php
danepowell: Does my comment make sense? I think the logic is correct. We should not run composer install if composer.json is absent.

danepowell:
This is ready for review (we even increased test coverage and MSI! 😮 )

src/Command/Pull/PullCommandBase.php
anavarre: Yes agreed, the current behavior to not run our logic when `composer.json` is missing is something we want to preserve.

What I was specifically trying to focus on here is the current issue we've discussed with the Trials use case which will have what we need (like a `composer.json` file) and also what we don't expect (a `vendor` dir).

I understand you have covered this here and think we should be good:

```
if (file_exists(Path::join($this->dir, 'vendor'))) {
      $this->io->note('Composer dependencies already installed. Skipping composer install.');
      return;
}
```

PR #1617, opened 2023-10-30
GL-1723: Added expires_at mandatory arg for codestudio token.

akashkska:
The ability to create personal access tokens without expiry was [deprecated](https://gitlab.com/gitlab-org/gitlab/-/issues/369122) in GitLab 15.4 and [removed](https://gitlab.com/gitlab-org/gitlab/-/issues/392855) in GitLab 16.0. In GitLab 16.0 and later, existing personal access tokens without an expiry date are automatically given an expiry date of 365 days.
Codestudio main branch is upgraded with gitlab 16.x version. hence this is breaking change. 
An optional variable in previous version https://archives.docs.gitlab.com/15.11/ee/api/project_access_tokens.html#create-a-project-access-token become mandatory in 16.x.x + versions https://docs.gitlab.com/ee/api/project_access_tokens.html#create-a-project-access-token.
Solution: As a part this PR added mandatory arg `expires_at` with valid expiry date.


src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -161,8 +161,13 @@ private function createProjectAccessToken(array $project, string $projectAccessT
       $this->checklist->completePreviousItem();
     }
     $this->checklist->addItem("Creating access token named <comment>$projectAccessTokenName</comment>");
+    $d = new \DateTime('tomorrow');
+    $formattedDate = $d->format('Y-m-d\TH:i:s\Z');
+    $format = "Y-m-d\TH:i:s\Z";
+    $dateTime = \DateTime::createFromFormat($format, $formattedDate);
```
danepowell, line 167:
```suggestion
    $dateTime = new \DateTime('tomorrow');
```

Why do you need four lines of code for this? Isn't `$d` identical to `$dateTime`?

danepowell (changes requested):

src/Command/CodeStudio/CodeStudioWizardCommand.php
akashkska: Actually implemented to get dateTime in specific format but it seems same is the default format. Hence change applied.

akashkska:
Hi @danepowell I don't see option to add label to this PR due to which one check was failing. This was breaking change for codestudio. will you please help out us to add label and merge this PR in main.

src/Command/CodeStudio/CodeStudioWizardCommand.php
```diff
@@ -161,8 +161,10 @@ private function createProjectAccessToken(array $project, string $projectAccessT
       $this->checklist->completePreviousItem();
     }
     $this->checklist->addItem("Creating access token named <comment>$projectAccessTokenName</comment>");
+    $dateTime = new \DateTime('tomorrow');
```
danepowell, line 164: This token needs to be long-lived since it's going to be stored as a CI/CD variable, right? Should we max out the lifetime (1 year)?

danepowell:
I'm going to go ahead and merge this. If you think the expiry time or anything else needs to be adjusted, let me know.

PR #1624, opened 2023-11-06
CLI-1192: Pin symfony/console

danepowell:


.github/dependabot.yml
```diff
@@ -2,6 +2,7 @@ version: 2
 updates:
   - package-ecosystem: "composer"
     directory: "/"
+    versioning-strategy: lockfile-only
```
danepowell, line 5: Even if they don't break tests and weren't responsible for this near-regression in the first place, the idea of unattended major updates makes me shudder. Plus it becomes difficult to pin dependencies.

PR #1638, opened 2023-11-22
CLI-1210: Drop tables instead of database

### LINKED ISSUE #1629: CLI-1210: `acli pull:db` uses the wrong dburl, ignoring ACLI_DB_HOST
**Describe the bug**

* When I do `acli pull:db ddevdemo.dev` it complains: "Unable to connect to local database using credentials mysql:://root:root@db/db. ERROR 1045
  (28000): Access denied for user 'root'@'192.168.229.2' (using password: YES)"

Note that the credentials should be `mysql://root:root@db/db` (it has two colons instead of one. (`mysql -uroot -proot -hdb --database=db` works fine)

Also, it's not using the correct database hostname (`db`) and instead looking up the IP of the calling container. (`db` has IP of `192.168.229.3` in this situation).  In this situation `192.168.229.3` is the IP of the calling container (ddev web container).

**To Reproduce**

Use `acli pull:db` with the environment variables above

**Expected behavior**

Database should be downloaded and imported. Instead, it fails with error shown.

**Additional context**

I'm working on updating DDEV's Acquia integration, and would like to use `acli pull:db` instead of the existing awkward approach, https://github.com/ddev/ddev/blob/49323379ad5cea0217a22fc0ee1d0725061022a0/pkg/ddevapp/dotddev_assets/providers/acquia.yaml.example#L46-L52

(It would be super nice if `acli pull:db --no-import` had a machine-readable output, and I would already be done, I could just grab the resultant dumpfile. There isn't a good way to do this is there?)

### LINKED ISSUE #1634: CLI-1218: Access Denied Error on pull:db using ACLI 2.20.1, Drupal 10.1.5 using Lando
**Describe the bug**
I'm running ACLI inside an 'acquia' recipe Lando box using Drupal 10.1.5 and ACLI 2.20.1 and receive the following error: 

```
gzip: stdout: Broken pipe

In PullCommandBase.php line 400:
                                                                                                                                                         
Unable to import local database. ERROR 1227 (42000) at line 6654: Access denied; you need (at least one of) the SUPER privilege(s) for this operation   
```                                                                                                                      
                                  

**To Reproduce**
Create a local environment using the Lando 'acquia' recipe with acli_version set to 2.20.1, php 8.1 and drush 10.  Run the command lando acli pull:database @[site].[env] 

The command will fetch the database but fail to import the database into the local database.

**Expected behavior**
I expect the acli command to pull down a database and import it without error

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 22
 - Browser N/A

**Additional context**
I AM able to direct download a DB backup from the Acquia Dashboard and use 'lando db-import' to directly import it.

danepowell:
Fixes #1629 
Fixes #1634 

Note, this has the side effect of no longer creating a db if it doesn't already exist.

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -189,7 +189,7 @@ abstract class PullCommandBase extends CommandBase {
     string $localFilepath,
     Closure $outputCallback = NULL
   ): void {
-    $this->dropLocalDatabase($databaseHost, $databaseUser, $databaseName, $databasePassword, $outputCallback);
+    $this->dropDbTables($databaseHost, $databaseUser, $databaseName, $databasePassword, $outputCallback);
     $this->createLocalDatabase($databaseHost, $databaseUser, $databaseName, $databasePassword, $outputCallback);
     $this->importDatabaseDump($localFilepath, $databaseHost, $databaseUser, $databaseName, $databasePassword, $outputCallback);
     $this->localMachineHelper->getFilesystem()->remove($localFilepath);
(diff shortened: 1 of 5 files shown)


danepowell:
I'm fine with this approach since it de-escalates the required permissions for these operations, which is always good. Cloud IDE should still work fine since required DBs are created via cron. Not sure about Lando.

PR #1643, opened 2023-11-28
CLI-1212: [auth:logout] remove secrets

danepowell:
@anavarre can you take this for a spin and provide feedback on the new UX and messaging?

I'm trying to give customers the flexibility to either "unset" the active key (so that they can re-use it by running `auth:login` again without having to re-input credentials) or to "delete" it entirely.

I'm not sure if "unset" and "delete" are entirely clear in this context. I'm open to feedback.

Delete credentials (the default option):

https://github.com/acquia/cli/assets/1984514/2ddb0fc9-de41-4f79-bc97-68bb775451cb

Unset credentials (interactive):

https://github.com/acquia/cli/assets/1984514/d658c899-02c5-4460-b9e4-6e7faa0dd413

Unset credentials (via `--no-delete` option)

https://github.com/acquia/cli/assets/1984514/030d3648-3bd2-450e-96b2-775d74d55f65

### DIFF
--- a/src/Command/Auth/AuthAcsfLoginCommand.php
+++ b/src/Command/Auth/AuthAcsfLoginCommand.php
@@ -11,7 +11,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
-#[AsCommand(name: 'auth:acsf-login', description: 'Register your Site Factory API key and secret to use API functionality')]
+#[AsCommand(name: 'auth:acsf-login', description: 'Register Acquia Site Factory API credentials')]
 final class AuthAcsfLoginCommand extends CommandBase {
 
   protected function configure(): void {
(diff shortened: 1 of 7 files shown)


danepowell:
The subscription isn't relevant here, since the key is attached to the user account, not a subscription.

I'll work to clean up the language.

I don't multiple prompts are necessary since the user clearly intended to do _something_ by running the command in the first place, and unsetting the credentials is easy to reverse by running auth:login again (I can mention that in the command output).

danepowell:
> Give the name of the subscription in bold

Oh I see what you mean though, the API key does have an associated label that we should use.

danepowell:
I think I've got this the way I like.

I found another significant bug in these commands, which is that they rely on isMachineAuthenticated when they really should only be considering keys in the datastore (since isMachineAuthenticated can also return true based on environment variables).

And that revealed that our tests were just mocking isMachineAuthenticated rather than properly mocking the Cloud datastore and credentials. I don't know how to fix that offhand, so unfortunately I have to remove those tests for now.

Issue #1661, opened 2024-01-13
CLI-1252: [New Command] To allow encrypt the text like Acquia Pipeline do


joshirohit100:
If we need any variable to be used in pipeline which is like secure, pipeline provides way to encrypt the variable from UI and then use that same in pipeline yml file.

This issue is to add a new command to support encrypt the text/string using ACLI.

Pipeline uses API `/api/v1/ci/encrypt` (got it from the network requets from Acquia Cloud)

danepowell:
I'm afraid we have to reject this request. Acquia already provides a Pipeline CLI with this functionality. While I'd love to see Pipeline CLI functionality integrated into Acquia CLI at some point, it would need to be done in a comprehensive way, not one ad-hoc command at a time, in order to keep the UX consistent, avoid confusion between the two tools, and ensure a stable architecture.

PR #1662, opened 2024-01-13
Adding new command to encrypt the variable like in pipeline.

joshirohit100:
Issue - https://github.com/acquia/cli/issues/1661

New Command -
```
./bin/acli pipeline:encrypt-var texttoencrypt
```

```
./bin/acli pipeline:encrypt-var --help
Description:
  Get the encrypted text to be used for pipeline

Usage:
  pipeline:encrypt-var <encrypting_string> <applicationUuid>
  pipeline:encrypt-var SOMESTRINGTOENCRYPT APPUUIDHERE

Arguments:
  encrypting_string     Text that to be encrypted
  applicationUuid       The Cloud Platform application UUID or alias (i.e. an application name optionally prefixed with the realm)
```

something same which is provided by pipeline client `encrypt` command

### DIFF
--- a/src/Command/Pipeline/PipelineEncryptVariables.php
+++ b/src/Command/Pipeline/PipelineEncryptVariables.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Acquia\Cli\Command\Pipeline;
+
+use Acquia\Cli\Attribute\RequireAuth;
+use Acquia\Cli\Command\CommandBase;
+use GuzzleHttp\Client as GuzzleClient;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[RequireAuth]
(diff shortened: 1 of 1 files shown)


joshirohit100:
If this is fine (approved), will add the tests

danepowell:
Why not just use the Pipeline CLI for this? I'm not happy about the idea of duplicating functionality between Acquia CLI and Pipeline CLI piecemeal. If we're going to merge Pipeline CLI into Acquia CLI, it needs to be part of a coordinated effort.

danepowell:
Our original vision was to eventually migrate all of the code in Pipeline CLI into Acquia CLI. I no longer think that's feasible since there's just so much of it.

This command is minimal by comparison because it's missing a lot of features of Pipeline CLI such as input validation, the ability to set custom endpoints, etc

Basically, this sets a precedent for the migration of a limited set of features from Pipeline CLI rather than the whole kit and kaboodle.

I'm okay with that but I want to see if @anavarre agrees. My worry is that the sort-of-but-not-quite duplication of functionality between Pipeline CLI and Acquia CLI will create confusion for customers. On the other hand, I think this migration can only _ever_ happen in this piecemeal fashion, and there is value in deprecating Pipeline CLI.

anavarre:
Yes, I'm afraid we don't want to set this precedent. Whatever relates to Pipelines CLI needs to be developed there until we hypothetically decide to merge both CLI together. The separation of concern is important IMHO.

PR #1685, opened 2024-02-16
CLI-1272: remote:drush fails with realm alias

danepowell:
I let refactoring get the best of me on this one, but it's for the better... not only does this allow several commands to be run with the standard environment prompt, I think this actually fixes a similar bug on ACSF environments with numeric prefixes (01dev). I confirmed that the expected directories in that case are `/var/www/site-php/profserv2.01dev`, not `/var/www/site-php/profserv2.dev` as our code and tests previously expected.

### DIFF
--- a/src/Command/Remote/DrushCommand.php
+++ b/src/Command/Remote/DrushCommand.php
@@ -14,13 +14,13 @@ use Symfony\Component\Console\Output\OutputInterface;
  * A command to proxy Drush commands on an environment using SSH.
  */
 #[RequireAuth]
-#[AsCommand(name: 'remote:drush', description: 'Run a Drush command remotely on a application\'s environment', aliases: ['drush', 'dr'])]
+#[AsCommand(name: 'remote:drush', description: 'Run a Drush command remotely on a Cloud Platform environment', aliases: ['drush', 'dr'])]
 final class DrushCommand extends SshBaseCommand {
 
   protected function configure(): void {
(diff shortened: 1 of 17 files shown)


src/Command/Remote/DrushCommand.php
```diff
@@ -14,28 +14,25 @@
...
       ->addUsage('myapp.dev -- status --fields=db-status');
   }
 
   protected function execute(InputInterface $input, OutputInterface $output): ?int {
-    $alias = $input->getArgument('alias');
-    $alias = $this->normalizeAlias($alias);
-    $alias = self::validateEnvironmentAlias($alias);
-    $environment = $this->getEnvironmentFromAliasArg($alias);
-
+    $environment = $this->determineEnvironment($input, $output);
```
danepowell, line 31: This is the meat of the fix, and has the side effect of changing the argument name from `alias` to `environmentId`. I'm 99% sure nobody will care or even notice.

anavarre:
This works great for me.

```
❯ acli remote:drush prod:REDACTED.prod -- status --field=drupal-version
/bin/bash: line 0: cd: /var/www/html/prod:REDACTED.prod/docroot: No such file or directory
9.0.6
❯ ./bin/acli remote:drush prod:REDACTED.prod -- status --field=drupal-version
9.0.6
```

anavarre (approved):

PR #1701, opened 2024-03-13
CLI-1292: Catch Cloud API type errors

danepowell:
@anavarre your opinion on messaging?

Typically a type error does indicate a problem with the application, i.e. some kind of broken configuration or environment. Sometimes it's a more general problem, which is actually true for CLI-1292. There's nothing wrong with the customer application in CLI-1292; the API will pretty much always return cron information that violates the published spec.

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -79,6 +79,10 @@ class ExceptionListener {
       }
     }
 
+    if ($error instanceof \TypeError && str_contains($error->getMessage(), 'AcquiaCloudApi\Response')) {
+      $newErrorMessage = 'Cloud API returned an unexpected data type; this could indicate a problem with your Cloud application.';
+    }
+
     $this->helpMessages[] = "You can find Acquia CLI documentation at https://docs.acquia.com/acquia-cli/";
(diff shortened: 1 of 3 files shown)


src/EventListener/ExceptionListener.php
```diff
@@ -79,6 +79,10 @@ public function onConsoleError(ConsoleErrorEvent $event): void {
       }
     }
 
+    if ($error instanceof \TypeError && str_contains($error->getMessage(), 'AcquiaCloudApi\Response')) {
```
anavarre, line 82: Seems like this line generates a lot of mutants. Can we make this more robust?

src/EventListener/ExceptionListener.php
```diff
@@ -79,6 +79,10 @@ public function onConsoleError(ConsoleErrorEvent $event): void {
...
+      $newErrorMessage = 'Cloud API returned an unexpected data type; this could indicate a problem with your Cloud application.';
```
anavarre, line 83: What about: "_Cloud Platform API returned an unexpected data type. This is not an issue with Acquia CLI but could indicate a problem with your Cloud Platform application._"?

PR #1715, opened 2024-03-29
CLI-1302: [api:applications:environment-create] databases must be array

danepowell:


src/Command/Api/ApiCommandHelper.php
```diff
@@ -75,10 +75,6 @@ private function addApiCommandParameters(array $schema, array $acquiaCloudSpec,
         elseif ($parameterSpecification['in'] === 'path') {
           $command->addPathParameter($parameterDefinition->getName(), $parameterSpecification);
         }
-        // @todo Remove this! It is a workaround for CLI-769.
-        elseif ($parameterSpecification['in'] === 'header') {
-          $command->addPostParameter($parameterDefinition->getName(), $parameterSpecification);
-        }
```
danepowell, line 81: I tested to confirm this is no longer necessary

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -21,7 +21,6 @@ class ApiCommandTest extends CommandTestBase {
 
   public function setUp(): void {
     parent::setUp();
-    $this->clientProphecy->addOption('headers', ['Accept' => 'application/json']);
```
danepowell, line 24: This is necessary to kill an escaped mutant. We need to check that this method is called, but it isn't actually called on every test case.

PR #1716, opened 2024-03-29
CLI-1294: [auth:logout] warning when key is invalid

danepowell:
@anavarre I'm trying to emphasize the self-help instructions for known exceptions by adding a bold "How to fix it" header, since it's otherwise easy to overlook. How do you feel about this?

Screenshot 2024-03-29 at 1 43 08 PM
Screenshot 2024-03-29 at 1 44 13 PM

For general exceptions where we can't offer specific help, you get just the second and third sentences about docs and support.

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -24,7 +24,7 @@ class Application extends \Symfony\Component\Console\Application {
     return $this->helpMessages;
   }
 
-  public function setHelpMessages(mixed $helpMessages): void {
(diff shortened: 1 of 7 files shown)


anavarre (approved):
This LGTM. The only thing that I'm unsure of is how it renders with different terminal schemes. I don't see a Phar attached so I can't try it out ATM. Can be dealt with in a different PR as needed.

PR #1725, opened 2024-04-11
CLI-1182: Improve Phar visibility for PRs

danepowell:


danepowell:
Oh... right... this involves building untrusted code, hence the pull_request target... but then we can't inject AWS credentials ☹️

danepowell:
Unfortunately there's no way to test this without the code being merged to HEAD first, so let's move this conversation to HEAD...

PR #1730, opened 2024-04-18
CLI-1317: api:base is not a valid command

danepowell:
For some reason people were actually running this command, and it throws typeerrors that get reported.


anavarre:
Works great.

```
❯ ./acli.phar api:base

In ApiBaseCommand.php line 92:

  api:base is not a valid command
```

PR #1741, opened 2024-05-23
Track environment provider in telemetry.

grasmash:
**Motivation**

We often don't know if ACLI is running in Code Studio, on local, or elsewhere.

**Proposed changes**

 Add an "env_provider" data point to telemetry, based on env variables, to determine env provider.


grasmash:
Ok, not sure why the mutation testing is failing. I tried to appease it, but to no avail.

PR #1764, opened 2024-07-08
GL-2753 : CS Wizard updated node version select values

PrakharJainS3:
**Motivation**

The current values provided in NODE_VERSION need to be update as for better onboarding flow to user on Node Project. Also the default value for the select needed to be changed.

**Testing steps**

Following steps can be taken to verify the changes:

- Run the following command `acli cs:wizard`.
- Please enter Cloud Key and Secret to login 
- Select Node_project in project type.
- In next select, the updated values will be visible for NODE version selection

### DIFF
--- a/src/Command/CodeStudio/CodeStudioWizardCommand.php
+++ b/src/Command/CodeStudio/CodeStudioWizardCommand.php
@@ -60,10 +60,10 @@ final class CodeStudioWizardCommand extends WizardCommandBase {
           break;
       case "Node_project":
         $nodeVersions = [
-            'NODE_version_18.17.1' => "18.17.1",
-            'NODE_version_20.5.1' => "20.5.1",
+            'NODE_version_18' => "18",
+            'NODE_version_20' => "20",
           ];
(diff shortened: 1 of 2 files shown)


anavarre:
Typically, Node versions are updated quite frequently upstream. It means you're gonna have to change the versions in ACLI equally often and get a new release out accordingly. Do we not have something more sustainable with less toil to envision?

PrakharJainS3:
That is what we are aiming for using these changes. So previously we had minor as well as patch version selected by user in the CLI, but now we will only provide the option to select the major version which changes in a long time.

anavarre:
Believe it or not, I actually looked at the diff the other way around. All good!

danepowell:
Looks good except for the failing mutation test. It's telling you that tests still pass when Node 18 is removed from the available options. Looks like you need to add/improve the test for Node 18.

danepowell:
Note that the coding standards on this project changed dramatically since you opened this PR. You'll need to rebase or merge on master, and it might be easier to just start a new PR.

PrakharJainS3:
Will open a new PR.

PR #1767, opened 2024-07-11
GL-2753 : CS Wizard updated node version select values

PrakharJainS3:
**Motivation**

The current values provided in NODE_VERSION need to be update as for better onboarding flow to user on Node Project. Also the default value for the select needed to be changed.

**Testing steps**

Following steps can be taken to verify the changes:

- Run the following command `acli cs:wizard`.
- Please enter Cloud Key and Secret to login 
- Select Node_project in project type.
- In next select, the updated values will be visible for NODE version selection


tests/phpunit/src/Commands/CodeStudio/CodeStudioWizardCommandTest.php
```diff
@@ -65,6 +65,8 @@ public function providerTestCommand(): array
                 [$this->getMockedGitLabProject($this->gitLabProjectId)],
                 // Inputs.
                 [
+                    0,
```
danepowell, line 68: The test cases are all providing invalid input, and since there are practically no assertions they were overlooked.

This fixes the inputs to at least be valid, but I'm not sure if the test cases are actually testing what they're supposed to...

PrakharJainS3: That is what I failed to understand, even when the inputs were wrong the output should not have contained the success messages which was asserted in the last.
I will probably keep this one to be worked upon in future.

danepowell: Yeah, I think this method only returns one line. Ideally you'd be checking the entire output:

```
        $output_strings = $this->getOutputStrings();
```

shriacquia:
@danepowell - would you be able to release this change? GL-2653 depends on this update.

danepowell:
https://github.com/acquia/cli/releases/tag/2.28.1

PR #1774, opened 2024-07-26
CLI-979: Consolidate update-check logic

### LINKED ISSUE #1403: CLI-979: Consolidate update-check logic
`CommandBase::hasUpdate()`, which runs on most commands _other_ than `self-update`, uses subtly different logic than the upstream self-update library to check whether updates are available. This leads to weird problems like #1402. We should consolidate this logic with the upstream library so prevent problems like this.

danepowell:
Fix #1403 

Waiting on https://github.com/consolidation/self-update/pull/34


danepowell:
I've added caching to the self-update library and created a new ticket to poll less often in the first place.

Just need to get tests passing now.

PR #733, opened 2021-12-03
Make `push:artifact` push to multiple git repos

grasmash:
**Proposed changes**

Add options to `acli push:artifact`:
* --destination-git-urls, an array of git urls to push to
* --destination-git-branch, the git branch to push to.

Allow same options to be defined in .acquia-cli.yml:
```
push:
  artifact:
    destination-git-url:
      - https://github.com/example/cli.git
      - https://github.com/example2/cli.git
    destination-git-branch: master
```

**Alternatives considered**

**Testing steps**

1. Follow the [contribution guide](https://github.com/acquia/cli/blob/master/CONTRIBUTING.md#building-and-testing) to set up your development environment.
2. Clear the kernel cache to pick up new and changed commands: `./bin/acli ckc`
3. (add specific steps for this pr)

**Merge requirements**
- [x] _Bug_, _enhancement_, or _breaking change_ label applied
- [ ] Manual testing by a reviewer

PR #1789, opened 2024-09-02
CLI-1389: push.artifact.destination_git_urls in acquia cli config

### LINKED ISSUE #1788: CLI-1389: "push:artifact" command has support for multiple git urls using acli config file but not supported in file.
acli `push:artifact` command supports multiple git urls and as per https://github.com/acquia/cli/blob/main/src/Command/Push/PushArtifactCommand.php#L164 config can be done using acli config file (acquic-cli.yml) but looks like code for this is not supported as AcquiaCliConfig only supports one key `cloud_app_uuid`

https://github.com/acquia/cli/blob/main/src/Config/AcquiaCliConfig.php#L23

This is my sample acli config file -
```
cloud_app_uuid: APPIDHERE
push:
  artifact:
    destination_git_urls:
      - GITURL1HERE
      - GITURL2HERE
```

joshirohit100:
Fix #1788 

Fix CLI-1389

### DIFF
--- a/src/Command/Push/PushArtifactCommand.php
+++ b/src/Command/Push/PushArtifactCommand.php
@@ -160,8 +160,8 @@ final class PushArtifactCommand extends CommandBase
         if ($envVar = getenv('ACLI_PUSH_ARTIFACT_DESTINATION_GIT_URLS')) {
             return explode(',', $envVar);
         }
-        if ($this->datastoreAcli->get('push.artifact.destination-git-urls')) {
-            return $this->datastoreAcli->get('push.artifact.destination-git-urls');
+        if ($this->datastoreAcli->get('push.artifact.destination_git_urls')) {
+            return $this->datastoreAcli->get('push.artifact.destination_git_urls');
         }
 
         return [$this->getAnyVcsUrl($applicationUuid)];
--- a/src/Config/AcquiaCliConfig.php
+++ b/src/Config/AcquiaCliConfig.php
@@ -19,10 +19,20 @@ class AcquiaCliConfig implements ConfigurationInterface
         $treeBuilder = new TreeBuilder('acquia_cli');
         $treeBuilder
             ->getRootNode()
-            ->children()
-            ->scalarNode('cloud_app_uuid')
-            ->end()
-            ->end();
+                ->children()
+                    ->scalarNode('cloud_app_uuid')->end()
+                    ->arrayNode('push')
+                        ->children()
+                            ->arrayNode('artifact')
+                                ->children()
+                                    ->arrayNode('destination_git_urls')
+                                        ->scalarPrototype()->end()
+                                    ->end()
+                                ->end()
+                            ->end()
+                        ->end()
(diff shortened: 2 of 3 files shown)


src/Command/Push/PushArtifactCommand.php
```diff
@@ -160,8 +160,8 @@ private function determineDestinationGitUrls(?string $applicationUuid): array
         if ($envVar = getenv('ACLI_PUSH_ARTIFACT_DESTINATION_GIT_URLS')) {
             return explode(',', $envVar);
         }
-        if ($this->datastoreAcli->get('push.artifact.destination-git-urls')) {
-            return $this->datastoreAcli->get('push.artifact.destination-git-urls');
+        if ($this->datastoreAcli->get('push.artifact.destination_git_urls')) {
```
joshirohit100, line 163: change from `destination-git-urls` to `destination_git_urls` is doing because for some reason (not sure why) if in https://github.com/acquia/cli/pull/1789/files#diff-e28b34b7f9414da3580d7e2274fd17c002beac308f6ab89aa7566f258198cd3aR28 use without underscrore that doesn't work but with underscore works.

So to me looks like somewhere this `-` is getting replaced with `_`

danepowell: We should be using underscores anyway since that's the convention (see `cloud_app_uuid`)

danepowell:
It looks like datastore support for Git URLs was added in https://github.com/acquia/cli/pull/733 but we weren't doing schema validation of the yaml store at that time. When we added the config tree validator I'm not sure why tests kept passing 🤷

danepowell:
This looks good but we need to figure out why tests didn't catch this to prevent it from regressing again.

Let me know if that's something you can figure out, or we can try to get this into an upcoming sprint.

joshirohit100:
I think command failing and test not failing because -

> For command when runs, container builds and thus acsf command factory or api command factory loads and loads other services - https://github.com/acquia/cli/blob/main/src/Command/Acsf/AcsfCommandFactory.php#L41 which loads the dataStore service. this further loads yamlstore service and that calls https://github.com/acquia/cli/blob/main/src/DataStore/Datastore.php#L69 and there this `processConfiguration` internally calls the treebuilder to validate the tree

and given in project repo, this acli config file already exists and thus validation fails while building the tree on container load / initialize

> In test run, this acli data store service is created from https://github.com/acquia/cli/blob/main/tests/phpunit/src/TestBase.php#L820 and as there is no acli config file exists, this passes and then there is no schema validation is done when something is set - https://github.com/acquia/cli/blob/main/tests/phpunit/src/Commands/Push/PushArtifactCommandTest.php#L128

danepowell:
@joshirohit100 thanks a lot for diving into that, I think you're right, the problem is we used `acliDatastore->set()` to establish test fixtures and this bypassed validation. We should have written the datastores as files and then reloaded the datstore/command to trigger validation. I'll fixed that in #1796

PR #1802, opened 2024-09-20
CLI-1397: Exception on local non-RSA SSH keys

danepowell:
**Motivation**

Fixes CLI-1397

**Proposed changes**

Remove the check for RSA keys

It looks like this was added in #1501 because we simply copied the code from the upstream library: https://github.com/violuke/rsa-ssh-key-fingerprint/blob/master/src/violuke/RsaSshKeyFingerprint/FingerprintGenerator.php

But I can't see any reason to restrict that code to RSA keys. You can run sha256 on any key 🤷 

**Alternatives considered**

**Testing steps**


danepowell:
Test failure is unrelated

PR #1839, opened 2025-01-03
CLI-1473: Pull default site if non-interactive.

charginghawk:
If your site has multiple databases, acli pull:db --no-interaction will not pull the default database, which is the logical default.

**Motivation**

Ran into this issue using `ddev pull acquia` on a site with multiple databases.

https://github.com/ddev/ddev/blob/master/pkg/ddevapp/dotddev_assets/providers/acquia.yaml#L53

It wasn't pulling the default, which I would expect.

**Proposed changes**

Use the default site when non-interactive.

**Alternatives considered**

**Testing steps**

Try `acli pull:db --no-interaction` on a site with multiple databases.

3. Check for regressions: (add specific steps for this pr)
4. Check new functionality: (add specific steps for this pr)

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -1411,7 +1411,7 @@ abstract class CommandBase extends Command implements LoggerAwareInterface
         if ($process->isSuccessful() && $sites) {
             if ($key = array_search('default', $sites, true)) {
                 unset($sites[$key]);
-                $sites = array_merge([0 => 'default'], $sites);
+                array_unshift($sites, 'default');
             }
             return $sites;
         }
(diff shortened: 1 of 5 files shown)


rfay:
The same is apparently true with pulling files. @danepowell We'd love your review of
* https://github.com/ddev/ddev/pull/6874

Is any of this behavior new, or did nobody ever pipe up before?

rfay:
First time I've ever seen an escaped mutant!

> Escaped Mutant for Mutator "DecrementInteger"

danepowell:
Thanks for the PR! I'd prefer to solve this slightly differently so that the question prompt reflects the default, e.g.:
```
 Choose a site [default]:
  [0] default
  [1] bar
  [2] foo
  [3] g
```

danepowell:
@rfay [mutation testing is amazing](https://dev.acquia.com/blog/better-tests-through-mutation-testing-or-killing-mutants-fun-and-profit)! In this case, it pointed out a potential bug with array_merge, if the array returned by `ls` contains a key less than zero. Not likely to happen, but better safe than sorry.

@charginghawk and @rfay please test and let me know if this works for you. You can download a dev build from one of the previous comments.

danepowell (approved):

PR #1842, opened 2025-01-17
CLI-1485: Add SSH URL in alias list

joshirohit100:
Adding  **SSH URL** column for the drush alias list.

**Impacted command** - `acli remote:aliases:list`

**Context** - We can ssh on server using `acli remote:drush ssh`. I needed to download some file (basically `scp`) from the server to local but I didnt have the details of the ssh of server to do the scp as SCP.

So I think it will be helpfull if drush alias list also provide this details so that this can be used if required

**Output before** -
```
+-----------------------------------+-------------------+--------------------------------------------+
| Application                       | Environment Alias      |   Environment UUID                           |
+-----------------------------------+-------------------+--------------------------------------------+
| My Application                    |  myalias.dev           |   MYENVUUID |
+-----------------------------------+-------------------+--------------------------------------------+
```

**Output after** 
```
+-----------------------------------+-------------------+----------------------------------
| Application    | Environment Alias | Environment UUID | SSH URL                                             |
+-----------------------------------+-------------------+--------------------------------------------+--
| My Application | myalias.dev       | MYENVUUID        | myapp.dev@myappdev.ssh.prod.acquia-sites.com  |
```

### DIFF
--- a/src/Command/Remote/AliasListCommand.php
+++ b/src/Command/Remote/AliasListCommand.php
@@ -26,6 +26,9 @@ final class AliasListCommand extends CommandBase
         $this->acceptApplicationUuid();
     }
 
+    /**
+     * @throws \Acquia\Cli\Exception\AcquiaCliException
+     */
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $acquiaCloudClient = $this->cloudApiClientService->getClient();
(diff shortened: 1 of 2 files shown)


danepowell:
This makes the table about 160 characters wide, which is way beyond what I'd consider healthy (80-132 chars). I'm also not sure what's so special about the SSH URL that it should be displayed here: people could easily want other info such as the Git URL, domain names, etc...

I can think of a few solutions, let me know what you prefer:

1. Move the application name out of the table since the command only takes a single application as an argument anyway. Put the name above the table.
2. Add help text after the table like `To get more information about a specific environment, run acli api:environments:find <alias>`, so people know how to get additional info

joshirohit100:
Use case for SSH URL was -

- I have access/subscription to an application in acquia cloud but don;t have application code on local
- I can ssh to acquia cloud using `acli drush`
- Then I needed to download some file to check/test from server and for that (`scp` command), I needed to know the SSH details (I could get that from acquia cloud but that required some steps)..

I think the missing part was as I didn;t know how to get the details after getting alias list.

So I think option-2 `To get more information about a specific environment, run acli api:environments:find <alias>` will be more helpful then

danepowell:
Sounds good, can you update the PR or open a new one to include that message?

joshirohit100:
Thanks! I have updated the MR.

Now this will look like -
```
+-----------------------------------+-------------------+-------------------------------
| Application    	   | Environment Alias   | Environment UUID                        |
+-----------------------------------+-------------------+--------------------------------
| Sample application 1 | devcloud2.dev       | 24-a47ac10b-58cc-4372-a567-0e02b2c3d470 |
+-----------------------------------+-------------------+-------------------------------

To get more information about a specific environment, run acli api:environments:find <alias>
```

danepowell:
I think you're right that having the SSH URL in the table could be useful. It's a bit incongruous with the name of the command (maybe it should be renamed `acli app:env-list` or something), but still useful.

Let me know if you approve.

joshirohit100:
Yes, I think changing the command name make sense because as per the current command remote:aliases:list,  app name is mandatory argument and command returns info about `Application name`, `env alias` and `env uuid`. And all of this info, `application name` looks redundant to me as this info is same for all the result + its already known.

Thus changing command name make sense to provide quick and useful info

PR #1878, opened 2025-07-09
CLI-1577: Update Cloud API spec

danepowell:
`./bin/acli` should run without errors and work with the entire API spec

To test, I'd like to see the `--help` output of every `api:` command before and after this change and make sure it's the same.

You can probably do this by running `self:make-docs` and comparing the output.


assets/acquia-spec.version
```diff
@@ -0,0 +1 @@
+3035ccd40bea9225a8b2a0fa421ffd9bde7b16da
```
danepowell, line 1: This just makes it easier to track what version of the spec we're using, since it's not encoded in acquia-spec.json itself

composer.json
```diff
@@ -57,7 +57,7 @@
         "zumba/amplitude-php": "^1.0.4"
     },
     "require-dev": {
-        "acquia/coding-standards": "^3.0.1",
+        "acquia/coding-standards": "^3.0.2",
```
danepowell, line 60: This isn't directly related to the PR; it was necessary to fix breaking changes in a recent release of PHPUnit

PR #1909, opened 2025-09-25
CLI-1628: Flaky environmenttransformer test error solved.

jigar-shah-acquia:
**Motivation**
Fixes [#CLI-1628](https://acquia.atlassian.net/browse/CLI-1628)

**Proposed changes**
- Updated test handling so that EnvironmentTransformerTest no longer crashes in parallel execution.
- Marked the test as @group serial to ensure stable CI runs on PHP 8.4.12.
- This change does not affect end users — it only stabilizes the CI process and release workflow.
- No manual updates required from end users.

**Alternatives considered**
NA

**Testing steps**

3. Check for regressions by running the full test suite on supported PHP versions (8.1, 8.2, 8.3, 8.4).
4. Confirm that no changes impact CLI user commands or runtime behavior.

**Reproduce the failure locally**

- Run paratest with the same flags used in CI:
```
vendor/bin/paratest --exclude-group serial --processes=2 --configuration=phpunit.xml.dist
```
- Specifically test EnvironmentTransformerTest.php with parallel workers to confirm crash behavior.
- Re-run above test with PHP 8.4.12 to confirm no crashes.
- Ensure other tests relying on environment tokens are unaffected. 
```
composer test
```


danepowell (changes requested):
I don't understand, why would this test class fail in parallel execution? It doesn't touch the filesystem, cache, environment variables, or have any other side effects that I can see.

danepowell:
This still doesn't make sense. The EnvironmentTransformer and EnvironmentTransformerTest are honestly the simplest classes and unit tests in our codebase, there's no reason they shouldn't be thread-safe.

I also can't reproduce locally or in GitHub Actions after dozens of runs. It only seems to happen on scheduled jobs. I'm wondering if this is a noisy neighbor issue. I'm going to adjust the schedule and see if that helps: https://github.com/acquia/cli/commit/1d304a54013694ac0636c518217317e1f23417d2

danepowell:
It's not just cron jobs where the failure occurs, it appears to be anywhere in CI: https://github.com/acquia/cli/actions/runs/18046727055

danepowell:
I suspect this is caused by the reflection used in test cases added a few months ago: https://github.com/paratestphp/paratest#caveats

Unfortunately I can only seem to reproduce this failure on `main`, especially on scheduled runs for some reason. I tried removing those test cases, we'll see if it helps: https://github.com/acquia/cli/actions/runs/18142465980/job/51636427798

If it does help, we should ban reflection in phpcs / phpstan.

If it doesn't help, we can restore those test cases and try `paratest --verbose` and/or opening an upstream issue.

danepowell:
I'm not confident this will fix the issue either in the short term or long term, but I'm running out of ideas so I guess let's give it a shot.

PR #1911, opened 2025-10-02
Remove Next ACMS from list of new applications

balsama:
**Motivation**

Next ACMS hasn't been maintained for over two years. I don't think we recommend this to customers anymore so we should remove it from the list of recommended installs.

**Proposed changes**

Removed the Next ACMS options from the `app:new` command.

**Alternatives considered**

We could choose to maintain the Next ACMS app, but that's would be a product decision. We should get someone from the Product org to approve this PR.

**Testing steps**

Run the `app:new` command. You should see two options. Prior to the PR, there were three options. I also added an assertion to `testNewDrupalCommand` that asserts this option is not there.

3. Check for regressions: (add specific steps for this pr)
4. Check new functionality: (add specific steps for this pr)

### DIFF
--- a/src/Command/App/NewCommand.php
+++ b/src/Command/App/NewCommand.php
@@ -59,15 +59,9 @@ final class NewCommand extends CommandBase
 
         $output->writeln('<info>Creating project. This may take a few minutes.</info>');
 
-        if ($project === 'acquia_next_acms') {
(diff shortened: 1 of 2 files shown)


danepowell (approved):
Thanks for the PR. I can tolerate a 0.02% reduction in coverage 😄

PR #1988, opened 2026-04-28
CLI-1778: success message immediately followed by a failure — confusing and misleading.

deepakmishra2:
This pull request improves error handling and user feedback in the database backup and pull process, and updates related tests for better reliability. The main changes ensure that missing backup download URLs are detected early, user messaging is clearer, and test assertions are more robust.

**Error handling improvements:**

* Added a check to verify that the backup download URL exists before attempting to download, and throw an `AcquiaCliException` with a clear message if it is missing.

**User feedback and messaging:**

* Refactored the backup wait logic to always display the "Database backup is ready!" message after a successful backup, regardless of how the callback is invoked.

**Testing improvements:**

* Updated the `testPullDatabasesOnDemandFail` test to use a try/catch block for more precise exception assertion, and added a check to ensure the success message is not shown when the backup fails.

**Code cleanup:**

* Removed an unused import of `React\EventLoop\Loop` from `PullCommandBase.php`.

### DIFF
--- a/src/Command/Pull/PullCommandBase.php
+++ b/src/Command/Pull/PullCommandBase.php
@@ -31,7 +31,6 @@ use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\TransferStats;
 use Psr\Http\Message\UriInterface;
 use Psr\Log\LoggerInterface;
-use React\EventLoop\Loop;
 use SelfUpdate\SelfUpdateManager;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\ArrayInput;
@@ -259,7 +258,9 @@ abstract class PullCommandBase extends CommandBase
         try {
             $codebaseUuid = self::getCodebaseUuid();
             if ($codebaseUuid) {
-                // Download the backup file directly from the provided URL.
+                if (!isset($backupResponse->links->download->href)) {
+                    throw new AcquiaCliException('Cloud API failed to provide a valid backup download URL. The backup may have failed on the server.');
+                }
                 $downloadUrl = $backupResponse->links->download->href;
                 $this->httpClient->request('GET', $downloadUrl, [
                     'progress' => static function (mixed $totalBytes, mixed $downloadedBytes) use (&$progress, $output): void {
@@ -395,15 +396,13 @@ abstract class PullCommandBase extends CommandBase
     protected function waitForBackup(string $notificationUuid, Client $acquiaCloudClient): void
     {
         $spinnerMessage = 'Waiting for database backup to complete...';
-        $successCallback = function (): void {
-            $this->output->writeln('');
-            $this->output->writeln('<info>Database backup is ready!</info>');
-        };
-        $success = $this->waitForNotificationToComplete($acquiaCloudClient, $notificationUuid, $spinnerMessage, $successCallback);
-        Loop::run();
+        $success = $this->waitForNotificationToComplete($acquiaCloudClient, $notificationUuid, $spinnerMessage, static function (): void {
+        });
         if (!$success) {
             throw new AcquiaCliException('Cloud API failed to create a backup');
         }
+        $this->output->writeln('');
+        $this->output->writeln('<info>Database backup is ready!</info>');
     }
 
     private function connectToLocalDatabase(string $dbHost, string $dbUser, string $dbName, string $dbPassword, ?callable $outputCallback = null): void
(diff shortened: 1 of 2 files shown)


tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
```diff
@@ -220,12 +220,18 @@ public function testPullDatabasesOnDemandFail(): void
         $this->setupPullDatabase(true, true, true, true, false, 0, true, false);
         $inputs = self::inputChooseEnvironment();
 
-        $this->expectException(AcquiaCliException::class);
-        $this->expectExceptionMessage('Cloud API failed to create a backup');
-        $this->executeCommand([
-            '--no-scripts' => true,
-            '--on-demand' => true,
-        ], $inputs);
+        try {
```
vishalkhode1: @deepakmishra2 This doesn't look correct to me. We shouldn't use try. /catch in PHPUnit tests, if we are expecting exception then we should add expectException (which was there earlier).

src/Command/Pull/PullCommandBase.php
```diff
@@ -395,15 +396,13 @@ private function getCodeabaseDatabaseBackups(string $siteId, string $environment
     protected function waitForBackup(string $notificationUuid, Client $acquiaCloudClient): void
     {
         $spinnerMessage = 'Waiting for database backup to complete...';
-        $successCallback = function (): void {
-            $this->output->writeln('');
-            $this->output->writeln('<info>Database backup is ready!</info>');
-        };
-        $success = $this->waitForNotificationToComplete($acquiaCloudClient, $notificationUuid, $spinnerMessage, $successCallback);
-        Loop::run();
```
vishalkhode1, line 403: @deepakmishra2 Why did we remove `Loop::run()`. How will now loop run if db update takes more time ?

vishalkhode1 (dismissed):

src/Command/Pull/PullCommandBase.php
deepakmishra2: The old code in PullCommandBase had two calls that started the event loop:                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                  
  1. LoopHelper::getLoopy(...) — which internally calls Loop::run() at line 62. This is the blocking poll loop that checks the Cloud API notification status every 5 seconds (with a 45-minute watchdog timeout).                                                                                 
  2. Loop::run() — called again right after getLoopy() returned in PullCommandBase.                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                  
  The second Loop::run() was redundant. By the time getLoopy() returns, the event loop has already run to completion — all timers were cancelled inside cancelTimers(), so the loop had no more work and exited. Calling Loop::run() again just immediately returned with nothing to do.

deepakmishra2: LoopHelper::getLoopy() still handles the entire polling lifecycle internally via Loop::run() on line 62. If the backup takes a long time, it will keep polling every 5 seconds up to the 45-minute timeout, same as before.

tests/phpunit/src/Commands/Pull/PullDatabaseCommandTest.php
deepakmishra2: makes sense, i'll revert this

src/Command/Pull/PullCommandBase.php
```diff
@@ -395,15 +396,13 @@ private function getCodeabaseDatabaseBackups(string $siteId, string $environment
...
+        $success = $this->waitForNotificationToComplete($acquiaCloudClient, $notificationUuid, $spinnerMessage, static function (): void {
+        });
         if (!$success) {
             throw new AcquiaCliException('Cloud API failed to create a backup');
         }
+        $this->output->writeln('');
```
vishalkhode1, line 404: @deepakmishra2 Removing `Loop::run()` is Ok, but we should remove this and instead use the callback function which was there earlier, instead of empty callback.

deepakmishra2: Ohh actually it was left intentionally because to provide a blank line after the spinner finishes. This was also there earlier on line 399

anujkaushal:
worked for me..
```
~/project $ ~/acli_dump/acli.phar pull:db <REDACTED> --on-demand
    ⌛ Creating an on-demand database(s) backup on Cloud Platform...

    ✔ Creating an on-demand database(s) backup on Cloud Platform
    ✔ Waiting for database backup to complete...
    ✔ Downloading <REDACTED> database copy from the Cloud Platform
    ✔ Importing <REDACTED> database download
```

```
~/project $ ~/acli_dump/acli.phar pull:db <REDACTED>
                                                                                                              
 [INFO] Using a database backup that is 0 hours old. Backup #<REDACTED> was created at Wed <REDACTED> 13:19:22 UTC 2026.
        You can view your backups here: https://cloud.acquia.com/a/environments/<REDACTED>/databases
        To generate a new backup, re-run this command with the --on-demand option.

    ✔ Downloading <REDACTED> database copy from the Cloud Platform
    ✔ Importing <REDACTED> database download
```

anujkaushal (approved):

Issue #1792, opened 2024-09-09
CLI-1392: The CLI keeps indicating that it requires an update


rpayanm:
**Describe the bug**
The CLI keeps indicating that it requires an update, but I already have the latest version. 
Additionally, I am unable to use it to obtain a backup.

**Desktop (please complete the following information):**
 - OS: Pop!_OS 22.04 LTS

Logs:
```
~ acli pull:database --no-import mydomain.prod mysite -vvv

Box Requirements Checker
========================

> Using PHP 8.2.23
> PHP is using the following php.ini file:
  /etc/php/8.2/cli/php.ini

> Checking Box requirements:
  ✔ The application requires a version matching "^8.1".
  ✔ The application requires the extension "json".
  ✔ The package "guzzlehttp/guzzle" requires the extension "json".
  ✔ The package "m4tthumphrey/php-gitlab-api" requires the extension "json".
  ✔ The package "ramsey/uuid" requires the extension "json".
  ✔ The package "zumba/amplitude-php" requires the extension "json".
  ✔ The package "composer/ca-bundle" requires the extension "openssl".
  ✔ The package "composer/ca-bundle" requires the extension "pcre".
  ✔ The package "vlucas/phpdotenv" requires the extension "pcre".
  ✔ The package "league/csv" requires the extension "filter".
  ✔ The package "m4tthumphrey/php-gitlab-api" requires the extension "xml".
  ✔ The package "marc-mabe/php-enum" requires the extension "reflection".
  ✔ The package "zumba/amplitude-php" requires the extension "curl".
  ✔ The package "laminas/laminas-servicemanager" conflicts with the extension "psr".
  ✔ The package "symfony/dependency-injection" conflicts with the extension "psr".
  ✔ The package "symfony/service-contracts" conflicts with the extension "psr".
  
                                                                                                                                              
 [OK] Your system is ready to run the application.                                                                                            
                                                                                                                                              

Acquia CLI version: 2.29.2
Acquia CLI 2.29.2 is available. Run acli self-update to update.
[debug] Using environment Prod ENV-UUID
PHP Warning:  Undefined property: stdClass::$user_name in phar:///usr/local/bin/acli/vendor/typhonius/acquia-php-sdk-v2/src/Response/DatabaseResponse.php on line 23

In ExceptionListener.php line 114:
                                                                                                                                             
  [Acquia\Cli\Exception\AcquiaCliException (1)]                                                                                              
[...]

PR #1993, opened 2026-05-18
CLI-1790: Fix nullable deprecation in 1790,1791,1792.

deepakmishra2:
**Motivation**

Fixes #1790, #1791, #1792


shubham-bansal96 (approved):

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -108,7 +109,7 @@ public function testArgumentsInteractionValidation(): void
                 'AH_SOMETHING',
                 'AH_SOMETHING',
             ]);
-        } catch (MissingInputException) {
+        } catch (MissingInputException | ValidatorException) {
```
vishalkhode1: @deepakmishra2 Ideally try / catch shouldn't be there in tests, but since it exists from earlier, so it's fine. Do we really need `MissingInputException` here. If not, then can we remove it from here and same in below.

vishalkhode1 (changes requested):

tests/phpunit/src/Commands/Api/ApiCommandTest.php
deepakmishra2: yes we can remove this class MissingInputException

vishalkhode1 (approved):

PR #2003, opened 2026-06-05
CLI-1810: run sanitize before cache clear.

deepakmishra2:
This pull request adds a new `--no-cache-clear` option to the `pull:database` command, allowing users to skip the `drush cache-rebuild` step after pulling the database. This is useful in scenarios where cache should not be cleared immediately, such as when pending update hooks need to run first. The logic for script execution is updated accordingly, and a new test is added to verify this behavior.

**New Feature:**
* Added a `--no-cache-clear` option to the `pull:database` command to allow skipping `drush cache-rebuild` after pulling the database. This is helpful when update hooks must run before cache is cleared.

**Script Execution Logic:**
* Updated the script execution order and logic in `PullDatabaseCommand` to respect the new `--no-cache-clear` flag, ensuring `drush sql-sanitize` still runs while `drush cache-rebuild` is skipped if the flag is set. [[1]](diffhunk://#diff-a155492c0c4429e1f0233ba4481844ec017cebff31943bc0cec6bd20db76b760R68-R80) [[2]](diffhunk://#diff-2b7771d88c66755dc00ee8758066c2b7ea6bc1bdb7f00c8c6dfbcbc96ac6cd55L2028-R2029)

**Testing:**
* Added a new test `testPullDatabaseNoCacheClear` to verify that when `--no-cache-clear` is used, `drush sql-sanitize` runs but `drush cache-rebuild` does not.
* Updated existing test order to match the new script execution sequence.

### DIFF
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -2078,6 +2078,29 @@ abstract class CommandBase extends Command implements LoggerAwareInterface
     /**
      * @throws \Acquia\Cli\Exception\AcquiaCliException
      */
+    protected function runDrushDatabaseUpdates(Closure $outputCallback, Checklist $checklist): void
+    {
+        if ($this->getDrushDatabaseConnectionStatus()) {
+            $checklist->addItem('Applying pending database updates via Drush');
+            $process = $this->localMachineHelper->execute([
+                'drush',
+                'updatedb',
+                '--yes',
+                '--no-interaction',
+                '--verbose',
+            ], $outputCallback, $this->dir, false);
+            if (!$process->isSuccessful()) {
+                throw new AcquiaCliException('Unable to apply database updates via Drush. {message}', ['message' => $process->getErrorOutput()]);
+            }
+            $checklist->completePreviousItem();
+        } else {
+            $this->logger->notice('Drush does not have an active database connection. Skipping updatedb');
+        }
+    }
+
+    /**
+     * @throws \Acquia\Cli\Exception\AcquiaCliException
+     */
     protected function runDrushSqlSanitize(Closure $outputCallback, Checklist $checklist): void
     {
         if ($this->getDrushDatabaseConnectionStatus()) {
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -2025,8 +2025,8 @@ abstract class CommandBase extends Command implements LoggerAwareInterface
     protected function executeAllScripts(Closure $outputCallback, Checklist $checklist): void
     {
         $this->runComposerScripts($outputCallback, $checklist);
-        $this->runDrushCacheClear($outputCallback, $checklist);
         $this->runDrushSqlSanitize($outputCallback, $checklist);
+        $this->runDrushCacheClear($outputCallback, $checklist);
     }
 
     /**
--- a/src/Command/CommandBase.php
+++ b/src/Command/CommandBase.php
@@ -2025,8 +2025,8 @@ abstract class CommandBase extends Command implements LoggerAwareInterface
     protected function executeAllScripts(Closure $outputCallback, Checklist $checklist): void
     {
         $this->runComposerScripts($outputCallback, $checklist);
-        $this->runDrushSqlSanitize($outputCallback, $checklist);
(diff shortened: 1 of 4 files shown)


src/Command/Pull/PullDatabaseCommand.php
```diff
@@ -59,15 +65,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
         $sourceEnvironment = $this->determineEnvironment($input, $output, true);
         $this->pullDatabase($input, $output, $sourceEnvironment, $onDemand, $noImport, $multipleDbs);
         if (!$noScripts) {
-            $this->runDrushCacheClear($this->getOutputCallback($output, $this->checklist), $this->checklist);
-            $this->runDrushSqlSanitize($this->getOutputCallback($output, $this->checklist), $this->checklist);
+            $outputCallback = $this->getOutputCallback($output, $this->checklist);
+            $this->runDrushSqlSanitize($outputCallback, $this->checklist);
+            if (!$noCacheClear) {
+                $this->runDrushCacheClear($outputCallback, $this->checklist);
+            }
```
anujkaushal: Why limit noCacheClear with noScript? it should act regardless of the script or noScript.

anujkaushal (changes requested):

src/Command/Pull/PullDatabaseCommand.php
deepakmishra2: this is if a user specifically run drush cr with --no-cache-clear. we are providing a new option to them

anujkaushal: But the same option will work if you keep it out of the if/else case. That way it can be used in `drush cr` without as well.

deepakmishra2: Ohh that makes sense if there is no option provided, we should keep that in a separate if. Let me refactor this

anujkaushal (approved):

src/Command/Pull/PullDatabaseCommand.php
```diff
@@ -65,9 +65,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
         $sourceEnvironment = $this->determineEnvironment($input, $output, true);
         $this->pullDatabase($input, $output, $sourceEnvironment, $onDemand, $noImport, $multipleDbs);
+        $outputCallback = $this->getOutputCallback($output, $this->checklist);
         if (!$noScripts) {
-            $this->runDrushCacheClear($this->getOutputCallback($output, $this->checklist), $this->checklist);
-            $this->runDrushSqlSanitize($this->getOutputCallback($output, $this->checklist), $this->checklist);
+            $this->runDrushDatabaseUpdates($outputCallback, $this->checklist);
```
vishalkhode1, line 70: @deepakmishra2 Order should be `Apply Pending Updates` → `Cache Clear` → `Sql Sanitize`.

src/Command/CommandBase.php
```diff
@@ -2025,8 +2025,8 @@ protected function getOutputCallback(OutputInterface $output, Checklist $checkli
     protected function executeAllScripts(Closure $outputCallback, Checklist $checklist): void
     {
         $this->runComposerScripts($outputCallback, $checklist);
-        $this->runDrushCacheClear($outputCallback, $checklist);
         $this->runDrushSqlSanitize($outputCallback, $checklist);
+        $this->runDrushCacheClear($outputCallback, $checklist);
```
vishalkhode1, line 2028: @deepakmishra2 I think this should also call `updb` and order should be `Apply Pending Updates` → `Cache Clear` → `Sql Sanitize`.

vishalkhode1 (changes requested):

src/Command/Pull/PullDatabaseCommand.php
deepakmishra2: i thought running sanitize will have impact on database tables like it might target email, username, password fields once ran and running drush cr at end would rebuild those cache. But had discussion that running sanitize after cache rebuild might not have any issue.

vishalkhode1 (approved):

PR #2011, opened 2026-06-24
CLI-1793: Consume Cloud API v3 OpenAPI spec in ACLI

deepakmishra2:
## Summary

- Wires the bundled `assets/acquia-v3-spec.json` (Cloud API v3 / MEO) into ACLI as `api:v3:*` commands
- Adds `V3ClientService` + `cloud.v3.connector_factory` — same credentials as v2, but base URI resolves via `ACLI_CLOUD_API_V3_BASE_URI` env var so v3 traffic can be pointed at dev/staging/prod gateway independently
- Fixes gateway URL and path-rewrite logic for MEO subscriptions (`AH_CODEBASE_UUID`-aware `PathRewriteConnector`)
- Adds `ApiV3CommandHelper` (reads `x-acquia-exposure.channels.cli.command` instead of legacy `x-cli-name`) and `ApiV3CommandFactory` to keep v2 and v3 command registration separate
- Adds mutation tests and spec structural sanity checks (`ApiV3SpecTest`)

## Test plan

- [ ] Set `ACLI_CLOUD_API_V3_BASE_URI=https://gateway.dev.api.acquia.io/v3` and run any `api:v3:*` command against the deployed dev gateway
- [ ] `acli list api:v3` shows commands generated from the bundled v3 spec
- [ ] Without `ACLI_CLOUD_API_V3_BASE_URI` set, v3 commands still resolve (falls back gracefully)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Command/Api/ApiCommandHelper.php
```diff
@@ -371,6 +371,42 @@ private function getCloudApiSpec(string $specFilePath): array
...
+    }
+
+    /**
+     * Extracts the stability level from an operation schema, or null if not declared.
+     * Override in subclasses that use a different spec convention (e.g. v3).
+     *
+     * @infection-ignore-all — protected→private is a false positive: PHP still dispatches
+     *   to the child's protected override via $this, so behaviour is identical.
+     */
+    protected function getSchemaStability(array $schema): ?string
```
vishalkhode1, line 393: @deepakmishra2 I don't think we need this function and also shouldSkipOperation as they are not doing anything.

vishalkhode1 (changes requested):

src/Command/Api/ApiCommandHelper.php
deepakmishra2: These exist in the base class purely as override hooks for V3, not because V2 needs them. Since ApiV3CommandHelper class extend this class we are making this method return null by default. Same for shouldSkipOperation

vishalkhode1 (approved):

PR #2013, opened 2026-06-29
CLI-1796: Productionize the v3 commands for CLI.

deepakmishra2:
This pull request updates the logic for filtering OpenAPI operations exposed to the CLI, adds corresponding tests, and improves the reproducibility of API spec builds. The main change is that operations can now be explicitly disabled for the CLI channel using the `enabled` flag, with tests added to verify this behavior. Additionally, the API spec build process now records the git commit hash used.

**CLI operation filtering improvements:**

* Updated `shouldSkipOperation` in `ApiV3CommandHelper.php` to skip operations that have `x-acquia-exposure.channels.cli.enabled` set to `false`, in addition to the existing audience filtering. The default is now to include operations unless explicitly disabled.
* Added unit tests in `ApiV3CommandHelperTest.php` to cover the new CLI channel filtering logic, including cases for `enabled: false`, `enabled: true`, and missing `enabled` (defaults to included).

**Build process improvements:**

* Updated the `composer.json` build scripts to write the git commit hash of the API spec to `assets/acquia-v3-spec.version` during the build process.
* Added the generated `assets/acquia-v3-spec.version` file, which contains the commit hash for traceability.


vishalkhode1:
I don't think we should pass [version=2](https://github.com/acquia/cli/blob/main/src/Command/Api/ApiBaseCommand.php#L115) for API v3.

vishalkhode1 (approved):

PR #2019, opened 2026-07-13
CLI-1827: Automate v3 API spec updates via daily cron PR

deepakmishra2:
This pull request automates and simplifies the process for updating the Acquia v3 OpenAPI spec. The main change is the introduction of a new GitHub Actions workflow that automatically downloads the latest spec, checks for changes, and creates a pull request if updates are found. Additionally, the manual update script in `composer.json` has been streamlined to fetch the spec directly from the remote source.

**Automation and CI/CD improvements:**

* Added a new workflow `.github/workflows/update-v3-spec.yml` that schedules a nightly job to download the latest Acquia v3 OpenAPI spec, checks for changes, and automatically creates and merges a pull request if updates are detected.

**Simplification of manual update process:**

* Updated the `update-acquia-v3-spec` script in `composer.json` to fetch the spec directly via `curl` from the remote source, removing the previous steps involving cloning the `api-specs` repo and running bundling commands.

These changes significantly reduce manual effort and ensure the spec stays up to date with minimal intervention.

### DIFF
--- a/.github/workflows/update-v3-spec.yml
+++ b/.github/workflows/update-v3-spec.yml
@@ -1,0 +1,50 @@
+name: Update Acquia v3 API spec
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
+  workflow_dispatch:
+
+jobs:
+  update-spec:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Download latest v3 spec
+        run: |
+          curl -fsSL https://api-specs-dev.dev.cicd.acquia.io/openapi.yaml \
+            | python3 -c "import sys,yaml,json; json.dump(yaml.safe_load(sys.stdin), sys.stdout)" \
+            > assets/acquia-v3-spec.json
+
+      - name: Check for changes
+        id: diff
+        run: |
+          if git diff --quiet assets/acquia-v3-spec.json; then
+            echo "changed=false" >> $GITHUB_OUTPUT
+          else
+            echo "changed=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create PR if spec changed
+        if: steps.diff.outputs.changed == 'true'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BRANCH="automated/update-acquia-v3-spec-${{ github.run_id }}"
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git checkout -b "$BRANCH"
+          git add assets/acquia-v3-spec.json
+          git commit -m "chore: update Acquia v3 OpenAPI spec"
+          git push origin "$BRANCH"
+          gh pr create \
+            --title "chore: update Acquia v3 OpenAPI spec" \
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -119,15 +119,6 @@
...
         ],
-        "update-acquia-v3-spec": [
-            "rm -rf var/api-specs",
-            "git clone --depth 1 git@github.com:acquia/api-specs.git var/api-specs",
-            "npm install --prefix var/api-specs @redocly/cli@latest --no-save --ignore-scripts",
-            "cd var/api-specs && PATH=$(pwd)/node_modules/.bin:$PATH make bundle",
-            "cd var/api-specs && npx --yes @redocly/cli@latest bundle --config redocly.bundle.yaml dist/openapi.yaml --dereferenced --ext json -o ../../assets/acquia-v3-spec.json",
-            "git -C var/api-specs rev-parse HEAD > assets/acquia-v3-spec.version",
-            "rm -rf var/api-specs"
-        ],
```
anujkaushal, line 130: Why remove this, lets keep it so we don't rely on github actions only.

.github/workflows/update-v3-spec.yml
```diff
@@ -0,0 +1,50 @@
...
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Download latest v3 spec
+        run: |
+          curl -fsSL https://api-specs-dev.dev.cicd.acquia.io/openapi.yaml \
```
anujkaushal, line 19: @deepakmishra2 this looks like dev URL?
can't we change to prod a variable which we can change later on without changing the code.

anujkaushal (changes requested):

.github/workflows/update-v3-spec.yml
deepakmishra2: stage and prod are still WIP so we will have to wait and then this will be changed to prod going forward, thats why this PR is going into feature branch and wont be released until prod is up and ready.

composer.json
deepakmishra2: we dont need this, the spec will be automatically be updated using github actions. This was added for dev team to update the spec locally which is not required now.

.github/workflows/update-v3-spec.yml
```diff
@@ -0,0 +1,50 @@
+name: Update Acquia v3 API spec
+
+on:
+  schedule:
+    - cron: '0 3 * * *'
```
deepakmishra2, line 5: Need to decide on this whether we should run this cron every day or every week.

deepakmishra2: So its decided we are going to run this daily, so no change is needed here

.github/workflows/update-v3-spec.yml
deepakmishra2, line 19: will move this to a environment variable. when we have the actual gateway url available we will move this to a separate helper method/class.

composer.json
deepakmishra2: discussion happened here - https://acquia.slack.com/archives/C0AR3UX6L01/p1781765030104799?thread_ts=1780916878.722269&cid=C0AR3UX6L01

anujkaushal (approved):

vishalkhode1 (approved):

PR #2025, opened 2026-07-20
CLI-1839: Guard against missing 'type' key in oneOf loop

deepakmishra2:
## Summary

- `castParamType` accessed `$type['type']` unconditionally inside the `oneOf` loop
- Any `oneOf` entry without an inline `type` (e.g. an unresolved `$ref`) triggered an `Undefined array key "type"` PHP Warning
- Fix skips entries that lack the `type` key via `array_key_exists` guard

## Related

- CLI-1839
- CLI-1799 / CLI-1745 (partial fix that resolved top-level `$ref`s but left this `oneOf` path unguarded)

## Test plan

- [ ] `testCastParamTypeSkipsOneOfEntriesWithoutTypeKey` — passes a `oneOf` spec with a typeless `$ref` entry alongside an `integer` entry; confirms the integer cast succeeds with no PHP warnings

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -696,22 +696,22 @@
         },
         {
             "name": "guzzlehttp/guzzle",
-            "version": "7.14.1",
+            "version": "7.15.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/guzzle/guzzle.git",
-                "reference": "6b1d2429a2c312474c523aa9017fba0c07b5f4a0"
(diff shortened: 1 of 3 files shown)


vishalkhode1 (approved):

composer.lock
deepakmishra2, line 1: The changes done was because composer audit --locked was failing due to these vulnerabilities and dependabot has not yet created a PR to fix this.

Image

PR #2028, opened 2026-07-22
CLI-1860: Updated response to include download links.

vishalkhode1:


deepakmishra2 (changes requested):
We should also add test for api:site-instances:database:backups with a _links.download fixture and asserts the download URL is preserved.

tests/phpunit/src/Commands/Api/ApiCommandTest.php
```diff
@@ -974,12 +976,14 @@ public function testTranslationsPathDoesNotMungeResponse(): void
...
-        // Verify _links was removed (munged by original 12-char check).
-        // If mutant (13-char check) exists, _links would remain and this assertion would fail.
+        // Verify _links was filtered (munged by original 12-char check) — still present but empty.
+        // If mutant (13-char check) exists, munging is skipped and _links retains its original keys.
         $this->assertIsArray($decoded);
         $this->assertCount(2, $decoded);
-        $this->assertArrayNotHasKey('_links', $decoded[0]);
-        $this->assertArrayNotHasKey('_links', $decoded[1]);
+        $this->assertArrayHasKey('_links', $decoded[0]);
+        $this->assertEmpty($decoded[0]['_links']);
```
deepakmishra2, line 984: and once we have the test we can change this assertion to assertArrayNotHasKey('_links', ...) for fixtures without a download key

PR #2031, opened 2026-07-30
IDE-5298: Improve federated auth error message to include AH_ORGANIZATION_UUID guidance

deepakmishra2:
This pull request improves the guidance provided to users when they encounter authentication errors related to Federated Authentication. The main change is that the help message is now dynamic: it checks if the `AH_ORGANIZATION_UUID` environment variable is already set and provides tailored instructions accordingly. The test suite has also been updated to cover this new behavior.

**Improvements to error handling and user guidance:**

* Updated the help messages in `ExceptionListener.php` to check for `AH_ORGANIZATION_UUID` in the environment and provide step-by-step instructions based on its presence. If set, users are prompted to run `acli auth:login`; if not, users are guided to export the UUID first, then run the login command.

**Test coverage enhancements:**

* Updated the expected help messages in `providerTestHelp` in `ExceptionListenerTest.php` to match the new instructions for when the UUID is not set.
* Added a new test, `testFederatedAuthHelpWithOrgUuidAlreadySet`, to verify that the correct help message is shown when `AH_ORGANIZATION_UUID` is already present in the environment.
* Added the `Group` attribute import to support test grouping.


deepakmishra2:
@copilot can you apply the suggestion given by you

PR #2044, opened 2026-09-07
CLI-1871: Add sync-skills GitHub Actions workflow.

deepakmishra2:
This pull request introduces a new GitHub Actions workflow, `sync-skills.yml`, to automate syncing ACLI skills with the `acquia-skills` repository. The workflow is triggered on releases and manual dispatch, and it handles version resolution, downloads the ACLI binary, generates command data, and dispatches a sync event to the target repository.

Workflow automation:

* Added `.github/workflows/sync-skills.yml` to automate syncing ACLI skills to the `acquia-skills` repository on release or manual trigger, including options for forced syncs.
* The workflow resolves the correct version tag, reads the Acquia spec version, downloads the appropriate ACLI binary, generates a JSON list of commands, and triggers a repository dispatch event with relevant payload data.

### DIFF
--- a/.github/workflows/sync-skills.yml
+++ b/.github/workflows/sync-skills.yml
@@ -1,0 +1,53 @@
+name: Sync ACLI skills
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      force:
+        description: 'Force PR even if no skill files changed'
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  sync-skills:
+    runs-on: ubuntu-24.04
+    name: Sync skills to acquia-skills repo
+    steps:
+      - name: Checkout release commit
+        uses: actions/checkout@v7
+
+      - name: Resolve version tag
+        id: version
+        run: |
+          if [ "${{ github.event_name }}" = "release" ]; then
+            echo "tag=${{ github.event.release.tag_name }}" >> $GITHUB_OUTPUT
+          else
+            echo "tag=${{ github.ref_name }}" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Read Acquia spec version
+        id: spec
+        run: echo "version=$(cat assets/acquia-spec.version)" >> $GITHUB_OUTPUT
+
+      - name: Download ACLI binary
+        run: |
+          curl -fsSL "https://github.com/acquia/cli/releases/download/${{ steps.version.outputs.tag }}/acli.phar" \
+            -o /usr/local/bin/acli
+          chmod +x /usr/local/bin/acli
+
+      - name: Generate commands JSON
+        run: acli list --format=json > /tmp/acli-commands.json
+
+      - name: Dispatch sync event to acquia-skills
+        env:
+          GH_TOKEN: ${{ secrets.ACQUIA_SKILLS_PAT }}
+        run: |
+          gh api repos/acquia/acquia-skills/dispatches \
+            --method POST \
+            --field event_type=sync-acli-skills \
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-skills.yml
```diff
@@ -0,0 +1,62 @@
...
+    runs-on: ubuntu-24.04
+    name: Sync skills to acquia-skills repo
+    steps:
+      - name: Checkout release commit
+        uses: actions/checkout@v7
+
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '8.2'
```
vishalkhode1: @deepakmishra2 Why do we need php here ?

.github/workflows/sync-skills.yml
```diff
@@ -0,0 +1,62 @@
...
+        id: spec
+        run: echo "version=$(cat assets/acquia-spec.version)" >> $GITHUB_OUTPUT
+
+      - name: Download ACLI binary
+        run: |
+          curl -fsSL "https://github.com/acquia/cli/releases/download/${{ steps.version.outputs.tag }}/acli.phar" \
+            -o "$RUNNER_TEMP/acli"
+          chmod +x "$RUNNER_TEMP/acli"
+          echo "$RUNNER_TEMP" >> "$GITHUB_PATH"
+      - name: Generate commands JSON
```
vishalkhode1: @deepakmishra2 What are we doing after adding the json in temp directory ?

.github/workflows/sync-skills.yml
```diff
@@ -0,0 +1,62 @@
...
+        with:
+          php-version: '8.2'
+          coverage: none
+      - name: Resolve version tag
+        id: version
+        run: |
+          if [ "${{ github.event_name }}" = "release" ]; then
+            echo "tag=${{ github.event.release.tag_name }}" >> $GITHUB_OUTPUT
+          else
+            echo "tag=${{ github.ref_name }}" >> $GITHUB_OUTPUT
```
vishalkhode1: @deepakmishra2 What's this `github.ref_name` here ? Isin't this should be `${{ inputs.tag }}` ?

.github/workflows/sync-skills.yml
```diff
@@ -0,0 +1,62 @@
...
+          fi
+
+      - name: Read Acquia spec version
```
vishalkhode1: @deepakmishra2 I don't think this is needed. We can directly run command and use in "Dispatch sync event to acquia-skills" step.

vishalkhode1 (changes requested):

.github/workflows/sync-skills.yml
deepakmishra2: https://github.com/acquia/cli/pull/2044#discussion_r3948427022

.github/workflows/sync-skills.yml
deepakmishra2: I think this was missed because the next step was to trigger and check the new command but we can do that as part of https://acquia.atlassian.net/browse/CLI-1870

deepakmishra2: Basically i thought of generating the command and pass it as payload for acquia-skills repo

deepakmishra2: I think i'm gonna change the plan to download the binary in acquia-skills repo directly instead of downloading it here

.github/workflows/sync-skills.yml
deepakmishra2: since we are removing acli binary from here i think php is no more needed now

vishalkhode1 (approved):

